PR #9, opened 2023-02-24
Reduce print statements

Femkemilene:
Debugging with print statements should be turned off by default.


ldxib2 (changes requested):
I like debugging statements

PVercoulen-CE (commented):
Great stuff

PR #26, opened 2023-05-23
Fttpower update instructions

Femkemilene:
I think I've covered all updates here. I will ask the two dissertation students to test the capacity updates in the near future. 

Femke


PVercoulen-CE (approved):
Thanks for adding. Minor remarks not worth not approving: The counters sometimes skip a number.

PR #32, opened 2023-06-23
Ftt cpat scenarios

PaulinaSchA:


Femkemilene:
Can you describe the pull request a bit better? There are over 5000 files changes, and unfortunately, github isn't great with showing diffs for large pull requests. This seems to include more scenarios. Do we want all scenarios on the main branch? How much does it grow in size?

PR #45, opened 2023-08-07
Major update of the FTT models

PVercoulen-CE:
- Updated input CSVs
- Removed all scenarios other than S0 to avoid having to download lots of CSV files. Users can generate their own files.
- Included RH's RK4 approach in each FTT module (therefore not entirely inline with EEIST results, but it follows the same trend)
- Removed unnecessary files.
- Updated the front-end manager
- Gamma tool does not initialise properly yet.


Femkemilene (changes requested):
As this is an omnibus change, I cannot do a standard review process and cannot see diffs, so I'm reviewing if
1.  The model can be run from the front-end and back-end, for all FTT models included
     * Yes, but there are an annoying amount of warnings of CSV files present and not include in VariableListing. Not a blocker, but would be some good clean-up task after the PR is approved. Files present are FR08 and TDA1.
2.  the output matches EEIST close enough, given the difference in RK4
     * There is an error with MWKA likely. The standalone model version does not fully out-regulate nuclear in Belgium, but stops one year short at 5k GWh/y. 
    * The total share of solar is quite a bit higher in the standalone version in 2050 (32M GWh vs 24M GWh in EEIST). I believe RK4 brings the fastest-growing technology up, but this is more than expected. Is this indeed due to RK4? And do we need to adjust gamma values to ensure solar is not growing too much after this adjustment?
3.  If the front-end manager does what I expect
     * The simple indicator now has a weird selection of variables. Two times curtailment, one of which seems to be defined as zero. I assume this was used for debugging, and should still be reverted to the old selection? There are now no variables listed for FTT:H and FTT:Tr in the simple overview; was that the case before?

Femkemilene:
I've linked this PR to one issue. Are there more issues that are solved from this PR?

PVercoulen-CE:
Response:

1. Indeed, some files are not used anymore. Is it preferable to delete them now or in the next iteration? Some variables/files will be needed at some point. TDA1 provides information on when historical fleet data ends. TDA2 provides information on when historical cost data ends. We can use this in the future to allow us to update the models in phases. I'll create a separate issue for this.
2. I'll check MWKA. RK4 improves accuracy which shows especially when there's lots of curvature (ergo it affects solar most). We definitely need to recalibrate. Other workstreams on substitution frequencies would help alot (not in terms of alignment with EEIST but in general), and tweaking LCOE by including WACC.
3. The settings for what is visible under the various view settings haven't been changed. I believe we should discontinue the various views and just include all variables as I expect this is the one that is predominantly used anyway.

Femkemilene:
2. If the TDA1 files are going to be used soonish, we might as well keep them. Given that this will become the production version, I think we should recalibrate gammas now, at least for the most important regions. Hope our next update will tone down solar a bit (as costs are now for 2020, whereas the model starts in 2019, so we're underestimating costs by a year). 
3. Agree; also saves the fuss of having to maintain multiple lists. Not sure why I'm seeing a weird selection, if this hasn't been changed.

PVercoulen-CE:
1. Actually, TDA1 can go. We got the "histend" parameter for each variable which makes TDA1/2 redundant.
2. Agreed on calibration. On solar, I think Ian has looked/is looking at different substitution frequencies and their effect. I think that might be the bigger effect. Also, we're missing grid expansion costs which would weigh more heavily on VREs. The question is whether that is something that should be part of the decision-making LCOE or whether it should be part of the price formation. 

Github is telling me that changes have been requested but I struggle to figure what changes specifically are requested. Can you elaborate?

rh-camecon (approved):
Looking at this, it seems to be bringing over many needed changes which happened in REFEREE. I'd like to accept these changes, and then I can compare the RK4 with my version to check for bugs, and also take over my changes to capacity additions and regulations so they can be tested by the team.

I also want to sort out the FTT-IH folder structure for OIS. 

As for issues with the frontend, I think best to accept the changes and then flag the issues with the frontend as 'issues'.

rh-camecon:
Let's wait until capacity additions and regulations are fixed before we redo the gamma values. I'm ready to make the pull request and for everyone to take a look, but I need this pull to be approved first, otherwise it's too tangled.

Femkemilene (approved):

Femkemilene:
In terms of changes requested, I wanted to make sure the MWKA fault was addressed. I've approved the change for now, as I don't want to hold up on further improvements.

In the future, we may want to make the distinction between a main branch and a production branch. That way, we may be able to work more easily together without approving changes to the main that may give odd results (as in: I don't trust we should have this much solar, and would like to mitigate the problem at least by having correct gamma values).

rh-camecon:
I agree, Femke, but part of this was caused by CE having a version which was quite ahead of Git, and me not realising Pim was talking over REFEREE, so I did all my changes in a branch from main which now causes merge conflicts. I think we will be working harder to avoid this in future, so we can do smaller pulls and the code can be diffed. I'm still trying to learn how to use Git so any more training or tips are appreciated!

Femkemilene:
This massive change should be a one-off. I think future merges will be much easier, but feel free to ask anything 🙂.

PR #51, opened 2023-08-16
Ensuring shares sum to 1

rh-camecon:
This update will make the following changes to how capacity or share additions are made in FTT stand alone, and how regulations are corrected for the stretching effect:

- Endogenous shares are calculated for every model following the substitutions made in the shares equation. 
- From here, endogenous capacity, generation, or useful energy demand is calculated. This difference depends on what the shares are shares of, e.g. in FTT: H market shares are calculated as shares of generation, so corrections should be made in generation space. 
- Share or capacity additions are corrected to the appropriate units for the given model, and scaled to the number of time steps. For this method, in the first year absolute differences from baseline will very closely match capacity additions (tested for TWSA). After the first year, the additions have stimulated demand and absolute differences will outstrip capacity additions.
- Note that the use of MWKA in FTT: P means the treatment is different. Setting MWKA to zero is now immediately effective, however all other values of MWKA may be adjusted due to the stretching effect. Nevertheless, even after setting >80% of capacity using MWKA, the model adjusts very well, and most targets are met after two years. This does not make a large impact of total capacity or total generation, which suggests the implementation works well. 
- As all corrections are done in capacity (or generation or UED) space, new shares are calculated using the new capacities. This ensures that if zero capacity has been mandated, there will be zero market shares. It ensures shares sum to 1 and prevents negative market shares. Any non-zero capacities will be subjected to the stretching effect, which will happen regardless of methodology.

All files are diffable so please critique the code and perform your own tests to confirm policy decisions are coming through as intended. If share/capacity additions every iteration becomes an issue, Cormac's method of only adding in the last iteration is also successful. Perhaps a switch for this should be a further development?

BONUS COMMIT

I have committed a launch.json file which, if debugging in vscode, will disable numba for you, so you do not need to comment any appearances of njit out if you want to debug the code. When debugging, select the Python: Debug File configuration.


SourceCode/Power/ftt_p_shares.py
```diff
@@ -145,59 +145,75 @@ def shares(dt, t, T_Scal, mewdt, mwdlt, mews_dt, metc_dt, mtcd_dt,
...
         MWKA_scalar = 1.0
 
-        dUkTK = (mwka[r, :, 0] - mewk_lag[r, :, 0])/no_it
+
+
+        # Correct for regulations using difference between endogenous capacity and capacity from last time step with endo shares
+            
+        dUkREG = -(endo_capacity - endo_shares*np.sum(mewk_dt[r, :, 0]))* isReg[r, :] 
+
+        #calculate capcity additions or subtractions after regulations, to prevent subtractions being too large and causing negatve shares.
```
Femkemilene, line 186:
```suggestion
        # calculate capacity additions or subtractions after regulations, to prevent subtractions being too large and causing negatve shares.
```

SourceCode/Industrial_Heat/ftt_fbt_main.py
```diff
@@ -381,39 +387,49 @@ def solve(data, time_lag, iter_lag, titles, histend, year, domain):#, #specs, co
...
+                dUk = dUkREG + dUkTK
+                dUtot_indirect = np.sum(dUk[:indirect_cut_off])
+                dUtot_direct = np.sum(dUk[indirect_cut_off:])
+
+                indirect_shares = (endo_ued[:indirect_cut_off] + dUk[:indirect_cut_off])/(np.sum(endo_ued[:indirect_cut_off])+dUtot_indirect)
+                direct_shares = (endo_ued[indirect_cut_off:] + dUk[indirect_cut_off:])/(np.sum(endo_ued[indirect_cut_off:])+dUtot_direct)
+                indirect_weighting = np.sum(endo_shares[:indirect_cut_off])
+                
+
+                # Calaculate changes to endogenous ued, and use to find new market shares
```
Femkemilene, line 425:
```suggestion
                # Calculate changes to endogenous ued, and use to find new market shares
```

SourceCode/Industrial_Heat/ftt_mtm_main.py
```diff
@@ -268,12 +267,13 @@ def solve(data, time_lag, iter_lag, titles, histend, year, domain):#, #specs, co
...
-        division = np.zeros([len(titles['RTI']), len(titles['ITTI'])])
-        division = divide((data_dt['IWK3'][:, :, 0] - data['IRG3'][:, :, 0]),
-                          data_dt['IRG3'][:, :, 0])
-        isReg = 0.5 + 0.5*np.tanh(2*1.25*division)
+        isReg = np.where(data['IRG3'][:, :, 0] > 0.0,
+                          (np.tanh(1 +
+                              (data_dt['IUD3'][:, :, 0] - data['IRG3'][:, :, 0]) 
+                                  / data['IRG3'][:, :, 0])),
+                          0.0)
         isReg[data['IRG3'][:, :, 0] == 0.0] = 1.0
```
Femkemilene, line 277: isReg is now a value between -1 and 1 everywhere. is that intentional? I believe it should be between 0 and 1, in a shape like 0.5 + 0.5 tanh(a + bx). If you put a=1.5 and b=10, you get quite reasonable behaviour.

SourceCode/Industrial_Heat/ftt_mtm_main.py
```diff
@@ -379,39 +381,50 @@ def solve(data, time_lag, iter_lag, titles, histend, year, domain):#, #specs, co
...
 
-                # Check that exogenous share changes add to zero
-                dUkTK = data['IXS3'][r, :, 0]/no_it
-                if (data['IXS3'][r, :, 0].sum() > 0.0):
-                    dUkTK[0] = dUkTK[0] - data['IXS3'][r, :, 0].sum()/no_it
+                # Convert exogenous share changes to capcity/useful energy demand. They do not need to sum to one.
+                dUkTK = data['IXS3'][r, :, 0]*Utot/no_it
+
+                #Check endogenous capacity plus additions for a single time step does not exceed regulated capacity.
+                #Convert ued to capcity for this check
```
Femkemilene, line 394:
```suggestion
                # Check endogenous capacity plus additions for a single time step does not exceed regulated capacity.
                # Convert ued to capcity for this check
```

SourceCode/Industrial_Heat/ftt_mtm_main.py
```diff
@@ -379,39 +381,50 @@ def solve(data, time_lag, iter_lag, titles, histend, year, domain):#, #specs, co
...
+
+                indirect_shares = (endo_ued[:indirect_cut_off] + dUk[:indirect_cut_off])/(np.sum(endo_ued[:indirect_cut_off])+dUtot_indirect)
+                direct_shares = (endo_ued[indirect_cut_off:] + dUk[indirect_cut_off:])/(np.sum(endo_ued[indirect_cut_off:])+dUtot_direct)
+                indirect_weighting = np.sum(endo_shares[:indirect_cut_off])
+                
 
-                # Convert to market shares and make sure sum is zero
-                # dSk = dUk/Utot - Uk dUtot/Utot^2  (Chain derivative)
-                dSk = np.divide(dUk, Utot) - data_dt['IWS3'][r, :, 0]*Utot*np.divide(dUtot, (Utot*Utot)) + dUkTK
+                # Calaculate changes to endogenous ued, and use to find new market shares
```
Femkemilene, line 419:
```suggestion
                # Calculate changes to endogenous ued, and use to find new market shares
```

SourceCode/Industrial_Heat/ftt_ois_main.py
```diff
@@ -367,6 +368,10 @@ def solve(data, time_lag, iter_lag, titles, histend, year, domain):#, #specs, co
                         #dSik[b1, b2] = S_i*S_k* (Aik*F[b1,b2]*Gijmax[b1] - Aki*F[b2,b1]*Gijmax[b2])*dt
                         #dSik[b2, b1] = -dSik[b1, b2]
 
+                #calculate temportary market shares and temporary capacity from endogenous results
```
Femkemilene, line 371:
```suggestion
                # calculate temporary market shares and temporary capacity from endogenous results
```

SourceCode/Industrial_Heat/ftt_ois_main.py
```diff
@@ -378,39 +383,50 @@ def solve(data, time_lag, iter_lag, titles, histend, year, domain):#, #specs, co
...
                 dSk = np.zeros((len(titles['ITTI'])))
                 dUk = np.zeros((len(titles['ITTI'])))
                 dUkTK = np.zeros((len(titles['ITTI'])))
                 dUkREG = np.zeros((len(titles['ITTI'])))
 
-                # Check that exogenous share changes add to zero
-                dUkTK = data['IXS5'][r, :, 0]/no_it
-                if (data['IXS5'][r, :, 0].sum() > 0.0):
-                    dUkTK[0] = dUkTK[0] - data['IXS5'][r, :, 0].sum()/no_it
+                # Convert exogenous share changes to capcity/useful energy demand. They do not need to sum to one.
```
Femkemilene, line 392:
```suggestion
                # Convert exogenous share changes to capacity/useful energy demand. They do not need to sum to one.
```

SourceCode/Power/ftt_p_shares.py
```diff
@@ -145,59 +145,75 @@ def shares(dt, t, T_Scal, mewdt, mwdlt, mews_dt, metc_dt, mtcd_dt,
         # Add in exogenous sales figures. These are blended with
         # endogenous result! Note that it's different from the
         # ExogSales specification!
```
Femkemilene, line 147:
```suggestion

```

Femkemilene (commented):
Did a quick review. Only important bit is the definition of isReg, which is now consistent (yay), but seems to range between -1 and 1, rather than 0 and 1.

SourceCode/Industrial_Heat/ftt_mtm_main.py
rh-camecon: isReg seems to be different in all versions, now I look at it. I did correct isReg to have the right variables in, but didn't look at the definitions. 

However, tanh( 1 + (C-R)/R) cannot become negative if both capacity C and regulation R are positive. It will go between 0 and 1. This is because the fraction part of this calculation cannot be greater than 1. 

Let's say capacity is 10 and the regulation is set to 20. Tanh(1 + (10-20)/20) = tanh(0.5) = 0.46
Let's say capacity is 10 and the regulation is set to 5. Tanh(1 + (10-5)/5) = tanh(2) = 0.96
For very, very small R, isreg will get closer and closer to 1.
For very, very large R relatve to C, isreg will be close to zero.

rh-camecon: Apologies for not catching the differences!

Femkemilene: I knew I was overlooking something (the [1+ ..] that ensures tanh doesn't drop below 1 ). 

I still think this is insufficiently strict when CAP>=REG (isReg = 0.76). We expect almost all new investments to be stopped at that point. The equation I fiddled out stops 95% of the investments at that point (0.5 + 0.5*tanh(1.5 + 10*[C-R]/R) ). The comments in the old code said 90% should be stopped. My equation is a bit ugly however, with quite random numbers.

rh-camecon: I'm happy to go with that equation if you have tested it, there is nothing wrong with ugly maths if it is ensuring the model works as intended.

I can't change my commit, but it looks like this will work for all the models (just checked it made sense for heat, where HREG is in shares for some reason) so if you want to put it in as a suggested change in all of them, I'm happy to add it to this pull!

Femkemilene:
I believe you can continue to commit in your branch, which automatically updates the pull request. May not have time till Friday.

SourceCode/Industrial_Heat/ftt_mtm_main.py
rh-camecon: Changes made.

Femkemilene (approved):

PR #53, opened 2023-09-25
fix various warnings

Femkemilene:
Fix divide by zero warning for regulation and deprecationwarning for daemon from latest change. There are various errors I didn't spot before the previous merge when running the industrial heat submodels (shares adding up to nan, rather than one). I don't think it's caused by the most recent merge, so I've not reverted.


rh-camecon (approved):
All changes look good. I will investigate industrial heat now; there are regions/sectors which do not have data so this may be the cause. Thanks, Femke!

PR #65, opened 2023-11-10
Clean up cost-curves code after question

Femkemilene:
Zoé, I have cleaned up the code a bit. Let me know if it's now sufficiently clear?


zoeberger (approved):
The code is clearer, due to better formatting/explanations/comments

PR #67, opened 2023-11-16
Changes from referee

Femkemilene:
This brings over bugs fixed in the REFEREE version at August 10 and 15. Additional changes are:
* A bug fix to avoid negative shares in Canada in FTT:Fr. The exogenous shares scaling factor went the wrong way around. Also corrected in FTT:Tr, no scaling correction present in Heat/Power 
* Bringing more of the code closer to PEP-8. Mostly resolving absence of spacing where required.

**Note**: this also brings over the IEA changes. Is that okay? Or is this data insufficiently transformed? Happy to cut out the data updates, and just include bug fixes

Description Jamie on the SVN server:

FTT-Fr code changes
    -Overhauled the LCOE calculation to account for average mileage to apply to the investment cost. This ensures all terms in the LCOE are done in per vkm terms ensuring a fair comparison between technologies
    -Added code for capacity additions ZWSA mirroring the treatment in FTT transport
    -Updated to Freight cost matrix 
    -Added fuel cost to EVs based on other EV columns 
    -Corrected mistake in efficiencies that divided EVs efficiency by 100
    -Updated the fuel efficiency of Electric small (vans) to reflect current European models  
    -Removed excessive gamma values now costs have been adjusted. 
    -Added ZWSA capacity additions for recent years using van and HGV EV stock data from IEA EV data explorer:  https://www.iea.org/data-and-statistics/data-product/global-ev-outlook-2023.


CamEconJP:
Thanks for bringing this across. I think we can keep the ZWSA in place for now to improve historical alignment but it should eventually be replaced by moving forward the last year of history.

@aileenlam28 may want to review this updated version of Freight code as she developed the original matlab code version that CE adapted to python

aileenlam28 (approved):

PR #71, opened 2023-12-01
Update VariableListing.csv

ldxib2:
Euros symbol causing encoding errors, changed to 'euros'


PVercoulen-CE (approved):
Noted, avoid use of non-alphanumerical characters

PR #74, opened 2023-12-08
Updated TEWW and variable name fixes

cormacmlynch:
TEWW is divided by 1000 - values previously too large. Updated VariableListing.csv with clearer names and units for some FTT-H and FTT-Tr variables. Front end variable names updated from current VariableListing.csv.


Femkemilene (approved):
I fixed an error that was caused by a couple of empty lines. I also included a new function that writes down more information when we have an error in reading in files. Nice that the front-end is now complete and FTT-Tr output is found there too.

PR #87, opened 2024-02-27
Ftt tr update cl patch

Femkemilene:
I've changed the convert masterfile_to_csv to:
* Generate non-existing files when called as part of run_file
* Overwrite csv files when run on its own. It will ask for confirmation to overwrite gamma values per your suggestion. 
I've tested it on FTT-P and FTT-Tr. It generates errors with the current version of FTT-P, but not with the version that Jasleen and Zoé are creating now. For tests:
* Can you record if the output is the same for key variables?
* Can you double check if the patch runs on your side too?


cormacmlynch:
Model output is verified for key variables (TEWW, TEWS, TEBC). The overwrite gamma values changes works well. 

Am I right in thinking that run_file.py should now generate all relevant csv files from the master sheets? I tried to run S0 FTT-Tr from the front end with the csv files not yet generated and came across an error:

image

cormacmlynch:
I also wonder if we could save even more space by removing the 'S0 only' variables from the S1 and S2 master sheets to make those files smaller?

Femkemilene:
I'll have a look at the error. Strange that running via frontend gives different results.. 

As for saving more space: we want to ensure that the S2 masterfile is the same as the FORTRAN file, so that we can easily copy between the two model versions.

Femkemilene:
@ldxib2   and me were chatting about this. Another problem here is the dependency on the celib package. We should have the information in the classification excel file to avoid this dependency. Ian will be working on rewriting this.

cormacmlynch:
Works for me for S0 FTT-Tr on front end now.

I think there might be an issue if you want to run an S2 scenario if the country CSVs don't exist yet? Its an easy fix to just create an S2 folder manually in 'Inputs' but this might not be clear to people. Is there a better work around?

Femkemilene:
The way I wrote it, it should work for S2 as well. What is the error you get with S2?

cormacmlynch:
Yes, works well if there is an empty folder but I guess this might need making clear?

Femkemilene:
There's two ways we could make this clear I believe:
* Add a FAQ in the documentation, and link to the FAQ in the readme file. This FAQ can also include explanation how variables should be added. Disadvantage is that people may not look for the FAQ
* Add a tooltip or simple explanation to the front end under the select scenarios. This will also need to explain that the baseline needs to be always selected. 

We'll need CE for the latter.

PR #96, opened 2024-03-06
Add installation instructions to readme

Femkemilene:
And tidy up the back_end file a bit: now displayed elapsed time without silly precision. Couldn't find how the greyed-out text can be adjusted


ldxib2 (approved):
Seems to run fine and instructions are clearer

PR #98, opened 2024-03-15
Ftt h update

AEdwards476:
Updates made to FTT-H so that the standalone version is the same as Cormac's recent updates in Fortran. Changes include updated data, updated variable names, functions in separate files and some changes to the way certain variables are calculated (e.g., HEWI). Note that during the update it was discovered that the implementation of tanh is different between the two models which effects isReg and shares. When isReg is turned off in both models the outputs are almost identical (<0.1% difference). Most up-to-date model on the svn is EEIST_FTT-H_Update


pv-camecon:
Just to check: The CSV files have been deleted because the masterfiles are now used directly instead?

Femkemilene:
Not quite. The code now automatically creates CSV files from the masterfiles when the model is run locally for the first time.

SourceCode/Power/ftt_p_rldc_v6.py
pv-camecon, line 1: I think this is an accidental commit due to being left-over from an older version of the main branch. Therefore this should be removed from the pull request.

pv-camecon:
The cost data have been updated. I presume this is based on more recent data from the Danish Energy Agency. What is the base year of the currency used (presumably Euro)? This is not per se relevant for the stand-alone tool, but it would be for the Fortran version. Most likely no base year is mentioned explicitly, then the year of publication can be assumed as the base year.

cormacmlynch:
@pv-camecon yes that's right, from the DEA's new costs which are 2020 EUR. I believe the base year is currently noted in the Fortran code but I'll add this as well to the FTT spreadsheet and the variable names.

cormacmlynch (approved):

PR #104, opened 2024-03-21
Changes (mostly to the frontend) for cross-platform use

### LINKED ISSUE #40: Unable to run model on MAC
Attached the error file when running the FTT model on Mac
[error_macJK (1).docx](https://github.com/cpmodel/FTT_StandAlone/files/11960024/error_macJK.1.docx)

ct-camecon:
# Overview

Code changes to run on non-Windows (e.g. macOS, Linux) platforms, with some minor tidying and supporting additions elsewhere.

Subject to (stress-)testing, this should close #40.

Almost all changes concern cross-platform, rather than Windows-specific, approaches to handling file and folder paths, for the most part using [`pathlib`](https://docs.python.org/3/library/pathlib.html).

Other changes:

* Switch to [`webbrowser`](https://docs.python.org/3/library/webbrowser.html) to open the frontend in the user's default browser, in place of the previous use of a Windows-specific system call: `os.system('start ...')`
* New 'requirements.txt', to `pip install` any dependencies prior to running
* Minor corrections, cleaning and tidying elsewhere

# Approach

Focus on getting the following to run:

1. `run_file.py`, which runs FTT from the terminal
2. `Backend_FTT.py`, which starts the browser-based GUI to run FTT and examine model results

**Note that the aim was to make changes that addressed runtime errors only**, rather than just blanket replace all instances. There are various functions in `Backend_FTT.py` that have been left alone for this reason, even if the corrections are (probably) safe and obvious.

Also note that the EXTRACT and GAMMA tabs have not been tested, pending further developments to those parts of the model and frontend.

Untouched functions, of which there are ten, have the following comment immediately below their function signature:
`# TODO: Still to review for cross-platform compatibility`

# Tests

Tests involved alternately running `python Backend_FTT.py` in:

* Windows Command Prompt, to check expected behaviour and that this behaviour remained after any code changes
* Windows Subsystem for Linux, running Ubuntu 20.04, to test on a non-Windows platform

Outputs and behaviour appear to be preserved between the two versions, as well as compared to the original fork (admittedly, to this non-FTT specialist).


pv-camecon:
@sgulati26 has been able to run the model using the front-end and it produces the same results as running it on a Windows machine. 

I'll leave the formal review to one of the Mac users.

ct-camecon:
@jp-camecon did some further tests on Windows to check for issues. Things *look* to be working as normal/before.

So we're just waiting for (at least) one Mac user to test this independently.

Femkemilene:
@jk694exe or @zoeberger : both of you have a Mac, right? Do you have time to test this?

cormacmlynch:
We just had someone with a Mac successfully run the model from the front-end in an EEIST2 FTT training session. Might still want a more formal review before merging though.

zoeberger:
When trying to run the Backend.py file from a Mac, there is an issue with the "start" command (cf picture). 
When I tried to use the equivalent for Mac (i.e. "open"), it created an issue with the thread object (cf picture).
Screenshot 2024-04-09 at 13 00 50

Femkemilene:
@zoeberger: in case you missed Chris' message.

pv-camecon (commented):
Given that this version has been tested by @sgulati26 and a few external students without leading to different results, I'm approving this pull request.

pv-camecon (approved):
*Previous comment was a misclick*

Given that this version has been tested by @sgulati26 and a few external students without leading to different results, I'm approving this pull request.

PR #109, opened 2024-03-27
Adding taxable fuel categories in FTT-Tr

cormacmlynch:
RTFT (fuel tax) is not technology dependent in the spreadsheets. In the Fortran model, we use an in-code converter to determine what vehicles are subject to the fuel tax. This was only semi-implemented in the Python code and it meant that EVs were being subjected to a fuel tax in the net zero scenario when they shouldn't be.

This should fix this for EVs, though I'm not sure what other vehicles should be exempt. Plug-in hybrids, hydrogen, CNG?

Also changed how the fuel cost component is displayed in the front end - now consistent with Fortran version and includes the fuel tax component.

### DIFF
--- a/SourceCode/Transport/ftt_tr_lcot.py
+++ b/SourceCode/Transport/ftt_tr_lcot.py
@@ -55,8 +55,10 @@ def get_lcot(data, titles, year):
 
     # Taxable categories for fuel - not all fuels subject to fuel tax
     tf = np.ones([len(titles['VTTI']), 1])
-    # Just EVs exempt at the moment
-    tf[18:21] = 0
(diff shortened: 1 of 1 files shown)


cormacmlynch:
Just to add too @Femkemilene, I couldn't replicate the error you mentioned today with creating the S2 CSVs. I only receive warnings for FR08 not being in the VariableListing file.

Femkemilene:
Would it not be cleaner to change the input values (both FORTRAN & here)? Less code, and it allows for more realistic scenarios to be made where some countries have a CNG tax exemption and others don't.

Femkemilene (approved):

PR #118, opened 2024-04-17
Small update of FTT-Freight

cormacmlynch:
As part of the EEIST 2 training, I have tidied the FTT-Freight code a little bit, making sure variable names are consistent with other FTTs and that policies work as we would expect them to. This update fixes some small bugs in the code, removes some variables that are not used, and improves the formatting in the masterfiles. The data has not been updated.


cormacmlynch:
To check from @AEdwards476's review: slightly unexpected values for ZEWI

PR #120, opened 2024-04-18
Remove /y from units

### LINKED ISSUE #100: Unable to sum across dimensions in the front end
For some variables, I am unable to find the sum across all regions or technologies in the front end. The 'Agg?' tickbox remains greyed out and unclickable.

image

Oddly enough, this works fine for some other variables.

Femkemilene:
So that they can be summed in front end. Closes #100.


pv-camecon (approved):
Good point. We should probably allow all variables to be summed regardless of whether it makes sense unit wise.

PR #123, opened 2024-04-19
old files in general replaced with new files from S0/FTT-P, costc error

ldxib2:
Bug was based on local files, corrected and works as expected


Femkemilene (approved):
Fixing this mistake has a very large impact: onshore wind now has more generation than solar. This is highly different from the FORTRAN version. The capacity factors are the same across the two models.

PR #128, opened 2024-04-29
Refactor input_functions.py

Femkemilene:
Rewrite input_functions so it has 3 fewer branches. Also ensures warnings for missing variables in VariableListing are only printed once.


cormacmlynch (changes requested):
Branch crashes when running S1, S2 scenarios.

cormacmlynch (approved):

PR #137, opened 2024-05-23
Update csv when masterfiles updated

Femkemilene:
This seeks to automatically update the csv files when the masterfiles is newer than the newest csv files. I think this will make it easier to update the model / switch between branches with different data. Should we try do finish this at the summer school?

Doesn't quite work yet, but nearly done:
* It seems to update all csv files when one masterfile is updated
* It's asking for confirmation to overwrite gamma values even when not

### DIFF
--- a/SourceCode/support/convert_masterfiles_to_csv.py
+++ b/SourceCode/support/convert_masterfiles_to_csv.py
@@ -5,10 +5,11 @@ Created on Wed Jan 16 17:10:10 2019
 This script extracts all the data from FTT excel sheets in the
 "/In/FTTAssumptions/[model]" folders and saves them in separate csv files.
 
-The user can select one or more scenarios to convert from the excel sheet. The 
-script overwrites when called directly, but does not when called from the run_file
-
-If running this code as a script, set the input files in the main bit of the code
+The user can select one or more scenarios to convert from the excel sheet:
+    1. If running this code as a script, set the input files in the final lines of the code
+    2. If the script is run at the start of the model run, it will automatically
+    detect which files to convert, based on available csv files and whether
+    the masterfile is newer than the csv files. 
 
 @author: MM and Femke Nijsse
 """
@@ -19,13 +20,12 @@ import sys
 
 import pandas as pd
 import numpy as np
-
 import datetime
 
 # Get the absolute path of the current script
 current_script_path = os.path.abspath(__file__)
 
-# Get the absolute path of the root directory (assuming the root directory is two levels up from the current script)
+# Get the absolute path of the root directory
 root_directory_path = os.path.dirname(os.path.dirname(os.path.dirname(current_script_path)))
 
 # Path to the 'support' directory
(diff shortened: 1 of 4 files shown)


Femkemilene:
@ldxib2: would you be able QA this one? If not, could you find somebody else to do it?
@cormacmlynch: I know you wanted to ensure behaviour that gamma values aren't overwritten. Under the condition that the Masterfile is newer than the newest csv, the default behaviour is to overwrite. Is that okay?
@pv-camecon: you said you would QA the model update in a week or 2. Without this PR, the model update requires you to manually run convert_masterfiles_to_csv.py. Otherwise, there will be an error. To avoid errors for others using the model, I think this PR needs to be tested & approved first.

SourceCode/support/convert_masterfiles_to_csv.py
```diff
@@ -115,24 +172,30 @@ def get_remaining_variables(vars_to_convert, out_dir, model, \
...
+    """
+    When the script is run as a script, and
     the gamma csv files already exist, confirm with
     the user if you want to overwrite, given that
     the user may not want to lose their calibrated gamma
-    values """
+    values 
+    """
     
     costvar_to_gam_dict = {"BTTC": "TGAM", "BHTC": "HGAM", "ZCET": "ZGAM"}
```
ldxib2, line 201: I thought a simple fix to add functionality for the Power module would be to update this list and the corresponding "GAMMA" cell in FTT_variables.csv. However, this has not worked and produces an error

ldxib2: Furthermore, the gamma csvs are still being overwritten for the non-power modules, even after inputting "n" to avoid overwrite

SourceCode/support/convert_masterfiles_to_csv.py
```diff
@@ -397,18 +398,25 @@ def get_model_classification(titles_path, variables_df):
...
     The main function to convert masterfiles to csv files. 
-    Depending on how you run it, it can have two types of behaviour:
+    Depending on how you run it, it can have three types of behaviour:
         a) If you run it as a script, it will overwrite files. It will
         ask for confirmation before overwriting the gamma values
-        b) If you call it from another function, it will only generate
-        new file if the files do not yet exist.
+        - If you call it from another function
+            b) It will generate files if files don't exist yet
+            c) It will overwrite files if the masterfiles are newer than the csv files
```
ldxib2, line 406: Functionality b) and c) are working. However, a) is not and is overwriting even despite check

ldxib2: issue seems to me to be in gamma_input_on_overwrite(out_dir, var, gamma_options) function. It's hard to tell what this is doing and I don't quite get the logic of the "gamma_options" argument?

SourceCode/support/convert_masterfiles_to_csv.py
```diff
@@ -157,14 +149,15 @@ def find_last_time_masterfile_modified(models, model, scen, dir_masterfiles):
     return master_time_mod > csv_time_mod
 
 def get_remaining_variables(vars_to_convert, out_dir, model, \
-                            var_dict, gamma_options, overwrite_existing):
+                            var_dict, gamma_options, overwrite_existing_csvs):
     """
-    Remove variables from list to convert if two conditions are met:
+    Remove variables from list to convert if either of two conditions are met:
         * Condition 1: csv files already exist or
         * Condition 2: The newest csv file is older than the masterfile
```
ldxib2, line 156: at first glance I thought list to convert are the variables we are going to update from the master. Am I wrong? If not then  surely we want to remove those variables that are newer than the master? Also if it is 'either' condition to be met, any file that exists will be removed, it doesn't matter if it is newer or older

ldxib2: Also there is a comment showing where condition 1 comes in but not 2

ldxib2 (commented):
Besides comments which I have put in separately, functionality has been tested and works including for the Power module. Code is also very readable

ldxib2 (approved):
Confirming review

PR #140, opened 2024-05-29
Adjusting FTT-Tr learning rates and turnover rates 

cormacmlynch:
The shares of EVs are too high in some regions. @aileenlam28 has discovered this could be due to different learning rates and turnover rates in the Python model. This pull request adjusts both of these values to match the MATLAB model. This includes a small difference to how the turnover rates are incorporated into the shares equation.

There are still some small differences between the models but this could be because of RK4 and different gamma values.


Femkemilene (approved):
Looks good. Just noting that TEWA is different from MEWA, which will make our life a bit more difficult when we create a single shares equation across models. In Power, MEWA already includes the timescales.

PR #144, opened 2024-06-04
Ftt ih update

rhayward1:
The version of FTT-IH in main is behind both in terms of data and bug fixes in the code. This version is consistent with the FTT Iter E3ME update, but most importantly contains corrected data, which ensures that the results of the model will make sense. 

This version should not cause any conflicts with main.


Femkemilene:
I fixed some indentation bugs, but can't test it properly as the frontend doesn't seem to work today on my work laptop. We've reworked the code so that, in general, the csv files are automatically updated from the masterfiles when running the code for the first time. Is there a reason that all the csv files are saved separately here still?

rhayward1:
> I fixed some indentation bugs, but can't test it properly as the frontend doesn't seem to work today on my work laptop. We've reworked the code so that, in general, the csv files are automatically updated from the masterfiles when running the code for the first time. Is there a reason that all the csv files are saved separately here still?

The masterfile for IH is out of date and doesn't have the newest data in it. The csvs are up to date. I created another issue asking for a code which could update the masterfile based on the csvs - I know this exists but I don't have a copy. Happy to sort this out today if you have that code!

pv-camecon:
Not sure what is going on with this pull request, but I'm unable to investigate the results through the front-end.

Femkemilene:
I fixed the indentation error again (this time it appeared in two models). When I try to run the model, I'm getting the following error:
image. @pv-camecon : is that similar to your error? That is, or can you run the model but not see the output?

Femkemilene:
Furthermore, in ftt_p_lcoe, the new gamma (1+gamma) is implemented. But this code doesn't yet have these new gamma values, right?

rhayward1:
This error happens because the IH models are missing from the FTT_Variables spreadsheet. I just added them all but I didn't commit as it would have conflicted with Main. I can commit the version I have and we can merge.

The new gamma value implementation shouldn't be in the update branch, only in the gamma automation branch, I can remove it. Sorry about that!

rhayward1:
Everything is now running if I run one IH model at a time, but I get an error if I try to run more than one. This seems to be caused by something in master files_to_csv.py.

SourceCode/initialise_csv_files.py
Femkemilene, line 1: @rhayward1: I think this has caused the error that only one model can be run at a time (not only IH). I don't quite understand the new code. If we need this, can you explain?

Femkemilene (changes requested):

SourceCode/initialise_csv_files.py
rhayward1: I believe @ldxib2 added this because the function was incorrectly splitting the module names if there was only one module run at a time. For example, when looking which module was the source of the error it would just say 'F' instead of 'FTT-P'.

Femkemilene: That's odd. I can't reproduce that error, so I've deleted the code as it definitely generates its own errors. 

I have changed the code slightly, so that it is more flexible. It can now understand input when you forget a comma (So FTT-H,FTT-Tr is now acceptable. This used to trigger an error).

ldxib2: Yes this was a change I made due to the error @rhayward1 described. New code looks better thanks

Femkemilene:
Sorry @rhayward1 for the slowness here. I'm getting another error running the model now (still examining if it's on my side)

A few requests based on visual code review
* There is still gamma value automation in this PR. For instance, the research_code.py file (which should probably get a more intuitive name in the gamma branch), and the code in titles_function. 
* MEWS seem to be saved in the Output folder. We shouldn't have output files on GitHub
* Can you copy the documentation into a Markdown file? I'm sure ChatGPT can do this quickly. Binary files such as Word documents are not trackable by Github, so it's difficult to track updates to the files.

rhayward1:
> Sorry @rhayward1 for the slowness here. I'm getting another error running the model now (still examining if it's on my side)
> 
> A few requests based on visual code review
> 
> * There is still gamma value automation in this PR. For instance, the research_code.py file (which should probably get a more intuitive name in the gamma branch), and the code in titles_function.
> * MEWS seem to be saved in the Output folder. We shouldn't have output files on GitHub
> * Can you copy the documentation into a Markdown file? I'm sure ChatGPT can do this quickly. Binary files such as Word documents are not trackable by Github, so it's difficult to track updates to the files.

Hi @Femkemilene, sorry, it's been a while so I'll need time to remember what's going on. I'm talking to Ian on Friday and will resolve this, then. Thanks!

rhayward1:
@Femkemilene I've merged main into this branch and sorted out any conflicts I can see. I've tried my best to remove any whiff of gamma value automation. 

As for converting the documentation into markdown - it contains a lot of figures and equations, and isn't code documentation but rather a report/record of how I created IH. I don't per se think it should be edited. Is markdown a good format for what is essentially a report? I mostly put it on GitHub to make it free and accessible, we can always take it off.

Femkemilene (approved):
The model now runs on my side, after I changed some code in convert_masterfile_to_csv. I'm a bit surprised the code ran for you. I'm approving, but before you merge it into main @rhayward1, could you check if it runs with these new changes on your side too?

rhayward1:
OIS was only pretending to run, and wasn't actually running. All models should be running now. Thanks.

PR #187, opened 2024-06-26
Update folder name for Linux compatibility (case-sensitive filepaths)

ct-camecon:
Amended a filepath in 'SourceCode/initialise_csv_files.py', when generating input CSV files:
* from: '_Masterfiles' (note the lowercase 'f')
* to: '_MasterFiles' (uppercase 'F', matching the case of the folder name in the repository)

Linux is case sensitive (whereas Windows and macOS aren't) so this difference matters. The lowercase 'f' failed to match any data and generate input CSV files. This led to a problem when running the model because the (non-existent) input files weren't read in, leading to missing arrays during solution.

Other than this now working on Linux, all other behaviour should be unchanged.

Otherwise, the file's been tidied slightly (stripped of trailing whitespace) and uses `pathlib` for the globbing.

### DIFF
--- a/SourceCode/initialise_csv_files.py
+++ b/SourceCode/initialise_csv_files.py
@@ -37,20 +37,22 @@ def get_masterfile(ftt_module, scenario):
     The filename that matches
     The middle bit of the file names for input to convert_masterfiles_to_cvs
     """
-
+    
+    # The * denotes the regionxtechnologies and last year updated part of the string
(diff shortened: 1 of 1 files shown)


Femkemilene (approved):
I can only test this on Windows (no problems). It looks like a straight improvement, so happy to approve.

ct-camecon:
Great. All fine with me (extra comments included). I'm not able to merge the pull request myself. Can I leave that to you/colleagues?

PR #190, opened 2024-07-08
Update ftt_p_main.py

rishids01:
data_dt gives the data at the previous time step. METC is the LCOE with all policies and gamma values included. METCX is the exognous LCOE (from E3ME output). We want to use endogenous LCOE values.


Femkemilene:
Looks good. What is the new global generation of solar in 2050? (MEWG)

rishids01:
If I am not wrong, the global generation of solar should be 22.71M GWH.

Femkemilene (approved):
Note: this brings the total solar production down again. This means the difference between E3ME-FTT and SA do need to be explored further.

PR #191, opened 2024-07-10
Rewrite carbon tax from MCOCX to REPP

rishids01:
Analysis files purged


Femkemilene:
It seems something went wrong in deleting the MCOCX files. 71 files, one for each country, should have been deleted, but only half or so are deleted. Maybe some bug with synching? Could you delete the other files as well before we start the full review?

PR #195, opened 2024-07-12
fix initialise_csv_files

Femkemilene:
The new automatic checks on updates in masterfiles revealed a pre-existing error, which has become a breaking error. That is:
* When running both a scenario with a corresponding masterfile and one without, the model_list for generating csv files becomes populated with Nones. Before, this stopped csv files from being created when they needed to. Now, it stops the code also for people with csv files already populated.


ldxib2 (approved):
Hotfix provides the proposed functionality and is commented appropriately

PR #216, opened 2024-08-08
Update ffp update manual

Femkemilene:
Include changes to how we want to update the model, partially based on our latest group meetings. This should make it more clear how costs are updated.


jp-camecon (changes requested):
Overall structure is fine. I have a few suggestions:

Historical generation - Do we have any guidance on mapping of data in the energy balances to FTT technologies is it all clearly one to one with the exception of Wind not being split? 

On differing start dates, I would change the To do from Option 1 and 2 to Simply TO DO. Both of these are desirable and not mutually exclusive (even if we update to more timely BNEF data, we should avoid hard coding the difference in start year of cost and shares into the code itself.

Costs - You mention checking against BNEF and Lazard for LCOE's. I would flag BNEF can be checked if you have access. For Lazard is there a source and is this open access?

Technical parameters - Do we have a source to reference here?

Femkemilene:
1. Apart from the wind split, we also need to make some assumptions on the CCS split if I recall correctly. I'm not sure what assumptions we made in the past. Globally, around 1.4% of coal-fired power plants have CCS according to ChatGPT, so we may need to verify that and assume a 98.6 / 1.4 split there. Not sure if IEA has this data directly, if not, could you update the manual when you've verified?
2. I've removed this completely, as Ian has coded this up in his data update. History end of BCET now needs updating.
3. Done
4. Added some references. Most of these updates need a different source each time however.

jp-camecon (approved):

PR #217, opened 2024-08-09
Further step, further corrections

Femkemilene:
We should update RERY, right? Some regions have zero values there now. How is this updated?

Rplace dead link to IRENA data


Femkemilene:
The other update is more up-to-date, so if we accept this one first, we can more merge the other update.

jp-camecon (approved):

PR #224, opened 2024-08-22
Correcting FTT-Tr costs

cormacmlynch:
Some of the country groupings for FTT-Tr costs were wrong. This update fixes this and corrects when learning starts in order to be consistent with the Fortran model. 

This pull request also contains a fix to how the fuel tax in FTT-Tr is implemented.

### DIFF
--- a/SourceCode/Transport/ftt_tr_main.py
+++ b/SourceCode/Transport/ftt_tr_main.py
@@ -54,8 +54,6 @@ from SourceCode.Transport.ftt_tr_survival import survival_function, add_new_cars
 #     return print("Hello")
 
 
-
-
 # %% main function
 # -----------------------------------------------------------------------------
 # ----------------------------- Main ------------------------------------------
@@ -101,7 +99,7 @@ def solve(data, time_lag, iter_lag, titles, histend, year, specs):
 
     sector = "tr_road_pass"
     sector_index = 0
-    sector_index = 15 #titles['FUTI'].index('16 Road Transport')
+    sector_index = 15  #titles['FUTI'].index('16 Road Transport')
 
     # Store fuel prices and convert to $2013/toe
     # It's actually in current$/toe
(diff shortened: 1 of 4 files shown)


SourceCode/Transport/ftt_tr_main.py
```diff
@@ -126,17 +126,28 @@ def solve(data, time_lag, iter_lag, titles, histend, year, specs):
...
+                data["TWWB"][0, veh, 0] = 0
+
+        # Save initial cost matrix (BTCLi from FORTRAN model)
+        data["BTCI"] = copy.deepcopy(data["BTTC"])
     elif year > 2012:
-        # Copy over TEVC values
+        # Copy over TEVC and TWWB values
         data['TEVC'] = copy.deepcopy(time_lag['TEVC'] )
+        data['TWWB'] = copy.deepcopy(time_lag['TWWB'] )
+        data["BTCI"] = copy.deepcopy(time_lag['BTCI'])
```
Femkemilene, line 143: We're getting rid of deepcopy in the code (np.copy does the same 100x faster).

Femkemilene:
The code runs okay. I'm doing some sense checks and getting unexpected results: In the S2 scenario, I now have more petrol mid in Belgium than in the baseline:

image

Not sure if I have some bleeding from another project in terms of input files. Are you getting the same results?

Femke

cormacmlynch:
I think there are two parts to this; before 2035 and after.

Before, this is, I think, a result of harsher taxes on the luxury variants in the S2 scenario which cause switching toward econ and mid diesel and particularly petrol.

After 2035 (in most European countries) there is quite a bit of switching happening from PHEV to petrol and diesel. Since all of these types are regulated out, this shouldn't be happening. So this is because of the regulation variable.

Important to note that EV shares are higher than baseline throughout.

Femkemilene (approved):

PR #225, opened 2024-08-23
Ftt tr costs correction

Femkemilene:


Femkemilene:
Does not work with all the other changes to main

PR #229, opened 2024-10-02
Turn of geothermal/hydro cost curves

### LINKED ISSUE #220: Turn off hydro cost curves
The current cost curves go up very steeply for large hydro. That is, while the starting CAPEX for Indonesia is 3300 \$ /kW, the start of the simulation already has a 10,000 $/kW CAPEX due to the cost curves, making the price unrealistically high (>$200 / MWh)  and at odds with published estimates of large hydro LCOE (typically under $100 / MWh). I see two options:

1. We turn it off now and redesign later.
2. We wait till we have time and redesign it. 

Given the long list of tasks, I propose we go for option 1.

Femkemilene:
For both technologies, the LCOE is now much more realistic. Geothermal LCOE is still quite high; hydropower LCOE is at times lower, and hydropower does grow currently in the US. The new IRENA data show a higher CAPEX in that region compared to what we have, so I think this will be resolved on its own. 

I think this should make the model's electricity price estimations in E3ME-FTT much more stable. 

This closes #220

### DIFF
--- a/SourceCode/Power/ftt_p_costc.py
+++ b/SourceCode/Power/ftt_p_costc.py
@@ -421,21 +421,21 @@ def cost_curves(BCET, BCSC, MEWD, MEWG, MEWL, MEPD, MERC, MRCL, RERY, MPTR, MRED
 
                         # BCET[r, j, 10] = MERCX[r, tech_to_resource[j], 0] * 2.0
 
-                # Increasing investment term type of limit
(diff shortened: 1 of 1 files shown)


ldxib2 (approved):
Simple changes just removing operations on techs where type of cost curve is set to 3.  This only affects Tidal, Hydro and Geothermal. Therfore, cost-curves turned off

PR #240, opened 2024-12-12
Updated cost data China + changing assumptions on boiler size

Femkemilene:
Update heat cost data using the [The Future of Heat Pumps in China](https://iea.blob.core.windows.net/assets/c099bae3-40f1-48cf-b905-64ff37755867/FutureofHeatpumpsinChina.pdf) report. 

I've taken the average of the two cities for most technologies (Figure 3.11). For coal, I used 3.12. The efficiencies of heat pumps are computed by the ratio of fuel prices between electricity and heat pumps. 

The costs are given as levelised costs in the report. I've added my backward calculation in the support repo, and would appreciate second pair of eyes (@AEdwards476) to check whether this is valid. The back-calculated LCOH is roughly the same as IEA. I've assumed costs are for 2021, as they're from a 2022 Chinese-language report. So I've decreased efficiency slightly, and increased costs for heat pumps to account for LBD.

Another question is whether we want to update the rest of the world based on this data. The only cost data we have directly is for Canada and EU countries if I remember correctly.


AEdwards476 (changes requested):
It all looks good to me. The only thing I noticed was that the O&M cost for air-air seems to be wrong in the masterfile. It should be 3.9 not 7.2. 
My only other note is that the District heating Operating costs (Other) for Harbin should be 0 (it is currently 0.07). This would then make the average 0 and therefore O&M cost also 0 (it is 0.2 in the masterfile). I don't think it makes a lot of difference in the grand scheme of things though.

Femkemilene:
I've updated the file in the support repo with additional comments for clarity.

I'm proposing we do the following for the other countries:
* Assume the same capacity of 8kW for each technology. Sits somewhere between the gas boiler and heat pump capacity. It seems like the gas boilers are often oversized, and that cost doesn't scale much with capacity. This means that dividing by their capacity isn't a fair comparison of cost. By using the same capacity, we get investment values closer to Florian LCOH values closer to IEA. 
* I propose we have two proxies: Western and Worldwide (average between western and China).  

I've implemented this in the TP branch, will bring it over here if possible.

Femkemilene:
Now updated all data with the new assumption (and fixed the error Alex spotted). Would be good to hear if @cormacmlynch agrees. Also updated gamma values.

cormacmlynch:
Interesting that the costs indeed don't really scale with capacity. I got some quotes from British Gas for a gas boiler in a standard semi-detached house and they were all in the range of £2,200-£3,000. Yet the capacities ranged from 18kw-30kw. So not too much difference. 

At the same time, perhaps we are overestimating our efficiencies for gas boilers? We assume 97% currently. I've had a look at some of the most popular gas condensing boilers (again in the UK) and most seem to be in the 92-94% range.

PR #241, opened 2024-12-17
Updates to battery + mandates

AmAkther:
Fixes an error with fix the battery (division by 0).

The new transport code is updated, very minimal overall changes. The code seems to work with the switch turned on/off. 

I've attempted to then update the Freight model and Heat model, however they don't seem to be reproducing the same values when the mandate switch is off.


SourceCode/Freight/ftt_fr_sales_and_mandate.py
```diff
@@ -0,0 +1,164 @@
+import numpy as np
+import copy
+green_indices=[12, 13, 18, 19]  # Indices for green technologies
+MANDATE_START_YEAR = 2025
+N_YEARS = 11
```
Femkemilene, line 5: In freight N_YEARS should be 16, as the mandate is slower than in the other two sectors.

SourceCode/Freight/ftt_fr_sales_and_mandate.py
```diff
@@ -0,0 +1,164 @@
...
+                    elif len(non_green_indices) > 0:
+                        sales_dt[r, non_green_indices, 0] = remaining_sales / len(non_green_indices)
+
+    # Step 5: Update capacity
+    adjusted_cap_growth_dt = sales_dt[:, :, 0] - base_eol
+    cap[:, :, 0] = cap_dt[:, :, 0] + adjusted_cap_growth_dt
+    cap[:, :, 0] = np.maximum(cap[:, :, 0], 0)
+
+    # Step 6: Update cumulative sales
+    sales_or_investment = copy.deepcopy(sales_or_investment_in)
```
Femkemilene, line 88: copy.deepcopy is rarely needed: np.copy does the same, but 100x faster.

SourceCode/Freight/ftt_fr_sales_and_mandate.py
```diff
@@ -0,0 +1,164 @@
...
+        for r in range(cap.shape[0]):
+            total_sales = np.sum(sales_dt[r, :, 0])
+            if total_sales > 0:
+                current_green = np.sum(sales_dt[r, green_indices, 0])
+                current_share = current_green / total_sales
+                
+                if current_share < mandate_share:
+                    target_green = total_sales * mandate_share
+                    
+                    # Keep existing global distribution logic
```
Femkemilene, line 53: Change the comment to: "#If green tech present in region, scale with region distribution". Only later are you taking the global values.

Femkemilene (commented):
Looks good :). Only comments for clarity and speed. Is this the final version or are you still making improvements?

AmAkther:
> Looks good :). Only comments for clarity and speed. Is this the final version or are you still making improvements?

Hi Femke,

Thanks for the comments :), I will keep trying to see what's going wrong in the Freight and Heat model however for now I'm deeply confused as to what's going incorrect.

PR #244, opened 2025-02-17
Eeist freight payload electricity price

Femkemilene:
Hi Amir,

Let's merge our two versions. You can commit your changes to the current repo, and then we'll try to merge. I've updated the electricity prices for most countries to take into account en-route charging per assumptions ICCT.

There are going to be some merge conflicts, as I've changed the mandates per country, rather than just an on/off switch. 

Do you have experience merging branches? Happy to help if you have not. 

Femke


AmAkther (approved):

AmAkther (changes requested):
Not fully checked, just reviewed those latest numbers.

PR #248, opened 2025-04-09
Vectorise lcoe

Femkemilene:
This PR contains a small correction to the original get_lcoe function, and a get_lcoe2, which is fully optimised (speed boost of almost 2x). Less of a speed bump again than I'd hoped, but does add up if we do this in the other FTT models too. 

Putting it in draft, so that the two functions can be easily compared. Before merging, we'll have to completely change to the new function. 

For review:

Can you make sure it's the same:
* For the different LCOE variables
* For different types of technologies (renewable / non-renewable)
* For a few different policies (subsidies f.i.). I've not tested the this yet on my side.


rishids01 (approved):
All looks good to me.

PR #250, opened 2025-04-11
Update ftt heat

Femkemilene:
I'm bringing over the major changes from the TP branch. Here the updated heat data and other changes I've made to that model

Questions for Alex:
1. Double check it doesn't throw errors & outputs are reasonable. I've not yet tested the S1/S2 files.
2. Check if the calculations for heating all make sense, and that the update manual is clear on how we got the new data. 

Questions for Amir
2. How should we implement the mandate? For the TP branch, we've done the simplest version now. That is, a switch to turn it on or off, which is overly simplistic. I see three options
2.1 Create a variable HGAM which would have a start and end date for each country
2.2 Create a variable HGAM which would have a mandate share in each year for each country.
2.3 Create two variables HGAM and HGAM_Sw. The first one has a share in each year for each country, the second the applicable technologies. 

I'm most keen on option 2, as linearly scaling might be too ambitious in the early years.


AmAkther:
For the mandate I propose a simpler solution that builds on the current implementation we have on the freight model. Here we would have five columns (titles) in a csv as such:
Region: Switch: Start Year: End year: Maximum share for mandate target.

We would predefine some curve that fits that information, right now we have a linear form however we may want the s curve adoption to maybe seem more realistic/feasible. The mandate adoption curve would then adjust from the csv information and the mandate code should just handle the rest. 

I think predefining each year for mandate % may be too time consuming? Not entirely certain however. If you wanted something mildly more specific/fine tuneable instead of every year we could have quarterly mandate percentage targets? It would therefore have  the same except the individuals would be able to specify exactly what they should have in the quarters of running the mandate code? I think this could capture a fairly massive amount of the information that would be caught in the yearly specific version(option 2.2). With regards to how the curve would fit this it would just be linear within each quarter.

Region: Switch: Start Year: End Year: Quartile 1 25% : Quartile 2 50% : Quartile 3 75 %: Maximum share

Femkemilene:
I don't fully understand your suggestion around quartiles. For instance, would that work if we implement mandates over a period that's not divisible by 4? Does this not make it difficult for people to understand how to implement this?

And do you think it would always make sense to follow a typical S-curve and hard-code that? Whether an S-curve is the right option depends on the maximum share. If we go from ~1 to 100%, then that would make sense. If instead, we'd like to move from 50 to 100%, it would make less sense. 

Cormac seemed most keen on a variant of option 2.3, in which the cost matrix would have the information about which techs are included in the mandate. This aligns with how we use the cost matrix in the power sector, but it's not a cost element, so might not be the most clear option..

AmAkther:
With regards to the division by four it wouldn't make much of a difference as it could simply factor it up or down by a year to get as close as possible, and then fit the linear curves in-between each point. However if time isn't a problem, as in people don't have an issue with inputting each year % by region then that will have the most accurate form and should be pursued for the highest accuracy. 

With Cormac's suggestion, I'm not entirely sure why we'd link it to the cost matrix?

Femkemilene:
We can either hard-code the information on which technologies are included in a mandate, or we can put this in a place where it can be modified, for instance in the cost matrix.

Femkemilene:
@AEdwards476 : could you check if it runs on your side too and works as expected? Happy to take care of the merge conflicts when it's approved.

AEdwards476 (approved):
All seems to be working well

PR #253, opened 2025-05-07
Ftt tr improvements

Femkemilene:
Speed up the LCOT code (like the power code). Add two new types of policy (base registration rate). 

Questions for Cormac:
* Shall I give the base registration rate a cryptic name? Do you think it'll be of interest for E3ME-FTT? Any suggestions?
* Is it alright to keep the carbon tax in the currency unit of FTT:Tr, or should I add a conversion?

Questions for Rishi:
* The code is correcting a bug wrt to the old version, in terms of the utility function. So, it should be almost the same, but not exactly. Can you double check it is? And that the various policies work as expected? I've tested the base registration rate and the carbon tax, but not the others.


rishids01:
You are absolutely right - there isn’t much difference between these outputs. The discrepancy I noticed was due to the printed output arrays from both models being different, which led me to misinterpret the yearly outputs.

Everything looks good on my end now.

Could you please clarify the sudden jumps in TEGC and TELC? Are these caused by the changes in the S0 files?

Femkemilene:
I'm not entirely sure. Do you see the jump in the new and old version? Do you see the jump in TETC as well? If not, it may be related to gamma values. Can you put a plot up?

rishids01:
TEGC_allcountries
TEGC_alltechs
The above ones are TEGC outputs from this branch, and from the main branch are herewith.
TEGC_main_allcountries
TEGC_main_alltechs
I don't see these changes in TETC.

cormacmlynch:
This is odd. I can't replicate the sudden jumps in TEGC (or TELC). I can't see an obvious cause in the code. 

image

About the registration tax rate, it will be easy enough to add to E3ME if its needed. TTRR maybe?

The carbon tax - sorry I don't see where the conversion from euros (REPPT) to dollars (units of FTT-Tr costs) taking place.

The other thing to note is that E3ME prices of middle distillates, electricity, and biofuels are read in (in TE3P) but they do not impact the fuel price in the cost matrix (which is what impacts the LCOT). No problem if this is intentional, but I wasn't sure what is intended to happen with the fuel prices.

cormacmlynch:
I checked an S2 scenario and it appears the levelised costs are far too high. I guess this is caused by either the road tax or the vehicle tax.

For reference, in the Fortran model, TETC S2 is around $0.67 /km.

image

Femkemilene:
I'm not reproducing that massive jump here weirdly enough. I do have a possibly unrealistic jump in price in my version too:

image

Femkemilene:
I've fixed the issue of the sudden jump (has omitted a more hidden conversion factor). Good catch. The costs are now 0.72, rather than 2.8. Still not the same as E3ME, but I don't think we've ever had alignment. The question is whether the alignment has worsened or improved. (Should be slightly improved, as I fixed a bug with the lifetime masking).

cormacmlynch (approved):
I get the same result now in S2.

PR #255, opened 2025-06-17
Use csv for classification_titles

Femkemilene:
Convert the classification file from Excel to csv to reduce the number of merge conflicts. This should make it easier to work on different model improvements in parallel. As always, please indicate if you cannot review within a couple of days, to stop bottle necks.


ldxib2 (approved):
Code makes sense and everything is functional.
What is the protocol for files like convert_excel_to_csv regarding relative paths? If you're working on an IDE with the working directory set as the FTT folder, this doesn't work but running with the utilities folder as working directory does. I presume the other models will be updated before this is integrated there?

Femkemilene:
Why do you think the other models need updating before this can be integrated?

Is there a better way to deal with paths to prevent this issue?

Femkemilene:
In main, those issues with power and transport are fixed. So, I'm merging now to the heat branch.

PR #256, opened 2025-06-19
Tidy up biofuel blending and update FTT:Tr

Femkemilene:
Clean up the biofuel blending code to be more similar to E3ME and faster. I've corrected TJET in the masterfiles. 

While reviewing you may encounter an unrelated bug: the phase-outs don't seem to work anymore, meaning that S2 is much less ambitious as the E3ME version. I'm investigating.


Femkemilene:
Just discovered that the code is not DRY, so will need to refactor the code. I also suspect there might be a logical error in the overall code for hybrids and PHEVs. Hope to resolve this before Monday, but not sure if I've got the time for it today.

Femkemilene:
The code is now refactored and fully sped up (FTT:Tr has a 15% increase in speed, excluding reading in files). The only correction I made to the code is around TJET, where hybrids and PHEVs now fully participate in the biofuel blending. So that means there are still two possible mistakes in the model (which are now flagged):
* The double counting of electricity vs fuel for hybrids and PHEVs
* The fact we have emissions for full-electric vehicles, which seemingly double counts with the power sector (right)? This one is easy to fix, but want to double check if I've not overlooked something.

SourceCode/Transport/emission_corrections.py
cormacmlynch, line 1: Very minor point but should we rename this to ftt_tr_emission_corrections.py to be consistent with naming conventions for other scripts?

Femkemilene:
I've now added the new data from E3ME, readded the Base registration rate, fixed the issue with lifetimes of bikes in China being 0.2 years. I've also adjusted the logic of the biofuel emissions factors with the new TJETs. Ready for review again.

PR #259, opened 2025-06-26
Vectorise early scrapping

Femkemilene:
Please only review if the two bits of code are the same. No need to review if the code is clear, as I doubt it's fully correct.


ldxib2:
key outputs checked and align identically with previous main branch

PR #265, opened 2025-08-01
Fix scenario names and timing module

Femkemilene:
With the sped-up code, it has become clear that the timing module didn't work (gave minutes, for a run of seconds). Also replace the naming of the scenarios from 'Test scenario', to something a bit more descriptive


AEdwards476:
Still giving wring timing for run, but the naming has been changed correctly

PR #269, opened 2025-08-08
Refactor regulation policies

Femkemilene:
Refactor regulatory policies. New code is more DRY and 30x faster than before. 

There are two logical changes:
* capacity now approaches MWKA more smoothly
* The regulation correction only works when demand is growing, in line with E3ME

I've kept the code to compare outputs, but will of course tidy up after review. For testing:
* MWKA is tested in baseline
* In the baseline, there are regulatory policies in power and heat, but not in transport
* Exogenous sales are off in S0 and S2, so need to be manually added for testing. 

MWKA is reverted back to its old consistent behaviour, so that there's no guessing what the policy does (there were 3 complicated conditions, which broke the logic of using exogenous capacity for partial data). 

Pim, feel free to delegate to someone else at CE!


Femkemilene:
Verbally, we agreed to refactor the code a bit more for consistency. This file will be split into two (exogenous sales and capacity).

Femkemilene:
Also: update requirements.txt as currencyconverter not in there

SourceCode/Power/ftt_p_main.py
```diff
@@ -27,25 +25,24 @@
...
+
+    - `rldc <ftt_p_rldc.html>`__
+        Residual load duration curves
+    - `dspch <ftt_p_dspch.html>`__
+        Dispatch of capcity
+    - `get_lcoe <ftt_p_lcoe.html>`__
+        Levelised cost calculation
+    - `survival_function <ftt_p_surv.html>`__
+        Calculate of scrappage, sales, tracking of age, and average efficiency.
+    - `cost_curves <ftt_p_costc.html>`__
```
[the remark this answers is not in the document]
Femkemilene: @copilot apply changes based on [this feedback](https://github.com/cpmodel/FTT_StandAlone/pull/269#discussion_r3147099804)

SourceCode/ftt_core/ftt_exogenous_capacity.py
```diff
@@ -0,0 +1,58 @@
...
+def exogenous_capacity(
+        exogenous_capacity, endo_capacity, dcap_other, regulation_cap,
+        t, no_it
+        ):
+    """
+    Calculate the change to endogenous capacity to reach exogenous capacities.
+    The goal is to move linearly to the new capacity over one year.
+    Regulation overrides exogenous capacity in case of a conflict.
+
+    When no_it = 20, initially you need to close 1/20th of the remaining
```
Xiaofan1999: This comment a bit confusing related to the change of capacity. A slightly more intuitive explanation of this  would be very helpful.
And could we add a brief explanation of the distinction between Exogenous capacity and Endogenous capacity?

SourceCode/ftt_core/ftt_exogenous_capacity.py
```diff
@@ -0,0 +1,58 @@
...
+    gap. In the next step, you need 1/20th of the original step, which is
+    1/19th of the remaining gap etc.
+    """
+    share_remaining_gap_to_close = 1 / (no_it - t + 1)
```
Xiaofan1999, line 32: The variable names `dcap_other`, `no_it`, and `share_remaining_gap_to_close` are a bit difficult to understand

SourceCode/ftt_core/ftt_exogenous_capacity.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Demand growth raises capacities while shares stay the same (stretching).
+    This extra capacity is not yet regulated. Correct for this underregulation.
+    """
+    cap_growth_from_stretching = endo_capacity - endo_shares * cap_sum_demand_dt
+    dcap_reg_corr = np.where(cap_growth_from_stretching > 0,
+                   -cap_growth_from_stretching * reg_constr,
+                   0)
+
+    return dcap_reg_corr
```
Xiaofan1999, line 62: After `dcap_reg_corr `is calculated in the `regulation_correction `function, will this value be used to further adjust the `dcap_exog_cap`

Femkemilene: Yes, it will be. I've renamed the dcap_other to dcap_reg_corr to clarify this

SourceCode/ftt_core/ftt_exogenous_capacity.py
Femkemilene: I've renamed dcap_other to dcap_reg_corr, and explained the two other variables in the docstring

SourceCode/ftt_core/ftt_exogenous_capacity.py
Femkemilene: Have improved the explanation here

SourceCode/ftt_core/ftt_exogenous_sales.py
```diff
@@ -0,0 +1,43 @@
...
+=========================================
+Contains the exogenous sales regulatory policy for FTT modules.
+
+Functions included:
+    - exogenous_sales
+        Calculate change in capacity from exogenous sales
+"""
+
+import numpy as np
+
```
Xiaofan1999, line 14: 1. Could you explain why exogenous sales are scaled down if they exceed 80%? Can we modify this threshold if needed? 
2. What is the intuitive meaning of `avg_lifetime `and `reg_overrides_exog`? 
3. I’m not entirely sure what the final output value represents, is it the total sales for a technology under this regulation, or the change in capacity for that iteration? 
4. The top comments are helpful to understand the policy goals or background, but the internal code logic is still a bit hard to follow. It would be helpful to add one-sentence comment for each code block to explain its purpose in both files. For instance, 
(1) In `ftt_exogenous_sales.py`, you could label blocks for "# Ensure exogenous targets do not exceed 80% of total demand for stability" "# Avoid applying policies in regions with zero demand" "# Let strict regulations override exogenous sales targets if they conflict" etc. 
(2) In `ftt_exogenous_capacity.py`,  "# Determine the capacity gap between current state and exogenous target" "# Correct under-regulation caused by overall market demand growth" etc.

(Perhaps because I am less familiar with energy market capacity, the logic feels a bit abstract to me; if these same coding structures described a regulation policy for corporate bond issuance limits, for example, the economic intuition would be much clearer, though I suspect the code is already quite clear to people with the right technical background)

Xiaofan1999 (approved):
Task 2
Power:
-> Exogenous capacity (MWKA) 
MWKA_DK: I increased the exogenous capacity for Solar PV between 2024 and 2026, total solar generation has a bit difference between two scenarios. In the frontend, the actual outputted capacity MEWK increased smoothly in S0 while increased significantly and peaked around 2031 in S_test, then started to decline as expected since no further interventions were made, but the overall level remains higher than in S0.
-> regulation (MEWR) 
MEWR_DK: I increased the regulation for Solar PV between 2024 and 2026, total solar generation has no difference between two scenarios. In the frontend, MEWK  is the same in both scenarios. I am not sure why as I expected S_test to be lower than S0 due to the regulation.

Heat:
-> Regulation (HREG) 
HREG_DK: Increased for gas between 2024 and 2026, total heat pumps has no difference between two scenarios. In the frontend, HEWK is the same in both scenarios (don’t know why).
-> exogenous sales(HWSA) 
HWSA_BE: Increased for oil between 2024 and 2026, total heat pumps has no difference between two scenarios. In the frontend, HEWK is the same in both scenarios.

Transport:
-> Regulation (TREG)   (couldn’t find TEWR so choose TREG as regulation)
TREG_DK: Increased for 1 Petrol Econ between 2024 and 2026, total electric cars has no difference between two scenarios. In the frontend, TEWK decreased sharply as expected in S_test compared to S0.
-> exogenous sales(TWSA) 
TWSA_DK: Increased for 1 Petrol Econ between 2024 and 2026, total electric cars has no difference between two scenarios. In the frontend, TEWK for Petrol Econ significantly increased in S_test compared to S0, peaking around 2026. After the intervention period, it continues to be slightly higher than in S0 through to 2050. 

Freight:
-> Regulation (ZREG)  (couldn’t find ZEWR so choose ZREG as regulation)
ZREG_DK: Increased for Petrol TWV between 2024 and 2026, number of electric trucks globally has no difference between two scenarios. In the frontend, ZEWK for Petrol TWV shows as 0 for both scenarios throughout the entire period. I am not sure why this is the case.
-> exogenous sales(ZWSA) 
ZWSA_DK: Increased for Petrol TWV between 2024 and 2026, number of electric trucks globally has no difference between two scenarios. In the frontend, ZEWK for Petrol TWV shows as 0 for both scenarios.

Task 3: I have reviewed the proposed changes on GitHub. As a fresher to the codebase, I didn't find any obvious logical flaws or errors during the code review itself.

Femkemilene:
@copilot : could you remove the old code, including the old model-specific files to implement this and the code comparing the old and new implementation with each other?

SourceCode/ftt_core/ftt_exogenous_sales.py
Femkemilene: 1. The exogenous sales come on top of the endogenous sales, so overall sales can start to exceed what is plausible at some point. 80% is not well-tested, but a decent first estimate of maximum sales. The code had a bug before, where this test did nothing. That is now fixed and tested. 
2. I've replaced the logic over average_lifetime (avg_lifetime) and instead compute the replacement rate per technology now, to better capture modules where lifetimes may vary. This is similar to how we do it in the sales function. I've explained reg_overrides_exog (a switch to turn of exogenous sales when it contradicts regulation)
3. I've clarified (it's both, because they are in the same unit)
4. I've added internal comments. 

You're absolutely right that the code was not super clear before. You're the perfect person to do these PRs, as you are not yet familiar with the code. 

In the below, you indicated that things do not yet work as you expected. When that happens, do not yet click 'approve', as we must be certain the code works before accepting. Could you recheck as discussed?

PR #274, opened 2025-08-14
Fix two bugs

Femkemilene:
First one is that the cost curve indexing was wrong. It started from 4 + K in FORTRAN, with K starting at 1, so from the 5th element. The python equivalent is 4:, not 3:. 

The other bug is more minor, but the capacity factor for CSP becomes non-sensical when the shares are 0, relying on the values from the previous technology. Now they're set to zero here, which means they'll be set to MWLO elsewhere.

### DIFF
--- a/SourceCode/Power/ftt_p_costc.py
+++ b/SourceCode/Power/ftt_p_costc.py
@@ -30,6 +30,7 @@ import numpy as np
 # ----------------------- fuel costs non-renewables ---------------------------
 # -----------------------------------------------------------------------------
 
+@profile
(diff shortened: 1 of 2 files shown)


cormacmlynch:
The CSP capacity factors are still quite erratic. Not sure if this is acceptable and an improvement on prior. For instance, France CSP capacity factor goes from 0.53 in 2018 to 0.28 in 2019.

Femkemilene:
The cause is that initially it´s taken from MWLO. Only when generation exceeds 0, are we starting to compute it endogenously. Realistic values are likely around .3, but cannot find direct data on it. Given that I´ll be replacing the data file, including wrong MWLO data later, I don´t think this is anything to worry about.

Issue #277, opened 2025-09-03
Correction for within-year capacity growth


Femkemilene:
In data, capacity is defined end-of-year, whereas generation is the total of the year. This leads to problems when capacity either grows or shrinks in a year, as the equation MEWL = MEWG / MEWK * conversion factor no longer works. 

In Fortran, a quick / temporary solution was implemented:

        !Adjust load factor from cost curve for renewables to account for new growth in capacity not operating for the full first year of operation
        WHERE (Svar == 1 .AND. MEWK>0) MEWL = ((MEWK - MEWI) * MEWL + (0.25 * MEWL * MEWI))/ MEWK

This is not ideal as
1. It redefines capacity factor, making it tough to compare with external sources
2. It only corrects for variable renewables
3. Possibly, this is leading to oscillatory behaviour in offshore / solar MEWG in the simulation period (as these oscillations are new, and I'm not sure what could cause it)

I believe the more elegant solution is to compute MEWG_t each time step, and compute MEWG cumulatively. This perserves the intuitive definition of MEWL, and should be easy to implement. If you agree Jamie, happy to implement in Python for testing. 

For the historical MWLO calculations, we need a different correction if we were to compute it directly from historical data. What we can do is simply compute MWLO[y] as MEWG[y] / (0.5 * MEWK[y-1] + 0.5 * MEWK[y] ).

PR #278, opened 2025-09-05
Update fttp attempt2

Femkemilene:
Last year, we updated FTT:P. This update got stalled, as I wanted to solve #277 first. Upon reflection, that issue is separate from this update, so we can proceed. Opening a new PR (and closing the old one), to make sure it works with all the recent changes to main. The updates here:

* Update costs to 2023 (BCET are the power sector costs)
* Update generation to 2022 (MEWG is generation)
* Change how we save the gamma values: the calibrated variation to costs based on historical uptake. This gamma value basically captures things like whether people prefer green energy, or prefer bigger car despite costs
* Change in the technologies included. We're removing a couple of quite niche technologies such as coal gasification, and adding categories for storage technologies. These storage technologies are not yet incorporated into the overall code base, so this is more of a preparation step. 

As the changes are quite significant, it would probably be good to check:
* Are there any nans in the output file?
* Are there any negative values in the storage costs (MSSC)?
* In the major regions, do we see solar and wind growing as expected? 
* Maybe copilot can come up with more things to check.

As this PR is more complicated, it might be a good idea to do it together with Francisco (I don't think he's on the repo yet).


SourceCode/Power/ftt_p_rldc.py
```diff
@@ -518,7 +501,15 @@ def rldc(data, time_lag, data_dt, year, titles):
         
         
         # Storage cost are overwritten here:
-        # Assumed levelised cost of storage: 0.20 EURO/kWh initially
+        if year >= histend["MSSC_histend"]:
```
Femkemilene, line 504: Should htis be larger than, rather than larger or equal?

SourceCode/Power/ftt_p_lcoe.py
```diff
@@ -171,7 +171,7 @@ def get_cost_component(base_cost, conversion_factor, mask):
     dlcoe = np.sum(npv_std, axis=2) / utility_sum
 
     # LCOE augmented with gamma values
-    lcoe_mu_gamma = lcoe_mu_all_policies + data['MGAM'][:, :, 0]
+    lcoe_mu_gamma = lcoe_mu_all_policies + bcet[:, :, c2ti['22 Gamma']]
```
Femkemilene, line 174: Should this be (1 + gamma)? Probably, right?

SourceCode/Power/ftt_p_rldc.py
ab-uoe: I believe this should be > to keep last historical year untouched until after the entire process and the block in line 650 overwrites it, but if i'm understanding this correctly, this will only affect outcomes if we did a historical data update in the same year such that year = histend year?

SourceCode/Power/ftt_p_lcoe.py
ab-uoe: should probably stay as is since it is additive and gamma is in $/MWh. I'm not sure how the automation process in Pull Request #279 would affect this though, I saw that it will be a value between -1 and 1?

ab-uoe (changes requested):
added comments to existing threads. 

Also line 264+ in ftt_p_costc.py, it seems that what get's summed to Biomass MEPD, which is MEWG[:, 8, 0] to MEWG[:, 11, 0], is not updated anymore? 8 is already ccgt+css, 9 ocgt, 10 ocgt +css.

ab-uoe (approved):

PR #279, opened 2025-09-05
Automate gamma values3

Femkemilene:
The gamma value automation script, which uses regulated simulated annealing to calculate gamma values. It runs all the selected models 5 years past their end year, and finds gamma values that minimise the difference in the 'rate of change' before and after the start of the simulation. 

To avoid finding local minima, we run the code 3 times. It's relatively slow, so consider setting no_it to 4 when you run the code. 

Is the code clear? I realise it's not the easiest algorithm, so hope my commenting is sufficient.

Testing:
a. Do the gamma values make sense? Check major technologies and major countries for all 
b. Are the gamma values comparable to manually determined gamma values?
c. Does the code work with different selections of modules? Other settings?

I'm building from the power sector update (which should be merged first), so only check the changes to automate_gamma_values


Femkemilene:
MGAMs are not stored on git, but on your local PC. To avoid deleting files you might want to keep, github retains these when switching branch. I'm not quite sure how to fix this: our solution to do away with the Excel files, move to big csv files (rather than chopped up ones) should lessen how annoying this is, but not fix it either. The solution here is likely to add a cleaner script, which is a separate issue, and should likely happen after we've converted the data inputs into big csv + parquet.

Do we want to preserve the gamma sliders? I would like to remove them altogether, as the algorithm should be good enough, and manual tweaking is not very scientific. I can try to remove that tab from the frontend; not sure if that's python or vue code.

Would you have time to synch this branch with main? Needs to happen before a merge anyway.

ab-uoe (commented):
Hi Femke, discussed w/ Ian some of his concerns with the automation algo. 

Points from him below: 

- The code runs successfully but the algorithm doesn’t work with many strong kinks in the shares projections; See Spain, Greece etc.
-  Works more effectively on countries with few technologies
- Still requires manual adjustment, but rate of change ROC, much less sensitive to changes
- Longer run, more iterations and smaller adjustment
- Normally 2 main techs need changing, gas/solar/on-shore

PR #282, opened 2025-09-11
Adding FTT-H2 to the master version

pv-camecon:
Pull request to include FTT-H2 into the master version.

There are a few oddities to sort out:
- CSC approach works in selecting the right technologies, but it fails to produce sensible prices. So far, we have relied on domestic LCOH2 estimates.
- Universal core FTT functions have been developed in parallel and need to be pulled into this version.
- Techno-economic cost data and technology classification may also change to include bio-based options.
- In the longer future, the CSC will be depreciated and replaced with another trade routine that is able to deal with bilateral information.


Femkemilene (changes requested):
Thanks Pim for this PR. Broadly speaking, there's 3 things I'd like to see
1. Having doc strings throughout. It's difficult to check the logic of the model when this is missing. A doc string should not only be a description of the variables, but also explain succintly what the functions do and why.
2. After we're happy with 1, it's time to check if we're happy with the logic of the code, and test the results are as expected.
3. Integrate the new core functions. Mostly, this will be adjusting this branch to align with the existing core functions, which are optimised and in some cases (sales) corrected. I can help here.

pv-camecon:
I'm closing this PR as it will be superseded by a separate one where FTT:H2 gets integrated into the current master and aligned with the existing structure within. There are also various areas of improvements required in this version of FTT:H2 which will need to be addressed.

pv-camecon:
Closed, see above

PR #289, opened 2025-12-17
Update environment

Femkemilene:
Hi Xiaofan,

Finally, as promised, a good first pull request for you. We're changing the installation instructions to be more in line with best practices. That is: moving away from cmd (which is Windows only), and using yml files with an up-to-date environment. Could you please:

1. Test if you can install the model with these instructions. Run all the four models using both the run_file method and the frontend.
2. Tell me if there are things I can do to make it more clear

Ideally, pull requests are reviewed within a few days, so that the submitter still remembers what they were thinking. If you need help with any of the individual steps, please do let me know and we can go over the PR together.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -140,7 +140,6 @@ dependencies:
       - click==8.3.1
       - fastapi==0.124.4
       - h11==0.16.0
-      - line-profiler==5.0.0
       - paste==3.10.1
       - pydantic==2.12.5
       - pydantic-core==2.41.5
@@ -149,4 +148,3 @@ dependencies:
       - typing-extensions==4.15.0
       - typing-inspection==0.4.2
       - uvicorn==0.38.0
-prefix: C:\Users\Work\anaconda3\envs\ftt


Xiaofan1999 (commented):
Hi Femke,

I followed the new installation instructions as a GitHub beginner, so some of the issues I encountered may be quite basic, but I hope they are still useful from a new-user perspective.

1. At the first step (git clone), Git was not recognised on my system. I installed Git first and then reopened Anaconda Prompt.

2. When running conda env create -f environment.yml, I initially encountered a file-not-found error because I was still in the previous directory. It would be better to remind beginner change path into the cloned FTT_StandAlone folder (where environment.yml is located).

3. During the environment creation, the process failed at the pip stage when building line-profiler, with an error indicating that “Microsoft Visual C++ 14.0 or greater is required”. I then installed the C++ components via Visual Studio Installer (“Desktop development with C++”), removed the partially created environment, and reran the installation following the instructions. The environment successfully installed.

4. After activating the ftt environment, double-clicking FTT_Stand_Alone_Launcher.cmd did not work and briefly showed “Python was not found”. Double-clicking Backend_FTT.py also did not start the server (browser showed “localhost refused to connect”). Typing `python Backend_FTT.py` from activated Anaconda Prompt worked, but the model could not run because the backend raised ImportError: Missing optional dependency 'openpyxl' (triggered by pandas.read_excel).

5. After installing openpyxl in the ftt environment, FTT-P, FTT-Tr and FTT-Fr can run from the frontend, but FTT-H can't (ValueError: Worksheet named 'HMAN' not found). I encountered the same situation when I run the models using run.py file.

Femkemilene:
That was a super useful review, thanks!

1. I've added a pre-requisite step of downloading Github desktop, which also installs git
2. I've added instructions to change directory to the FTT folder first
3. ChatGPT explains this is due to the fact that line-profiler doesn't have Windows Wheels yet with python 3.14. That's supposed to happen soon, but to make sure the instructions work now, I've deleted line-profiler from the yml file. 
4. I've added instructions about adding Python to your path + added openpyxl to the yml file. I'm not entirely sure what the second error is, but I'm considering these types of errors out of scope (I have idea how to fix it). Cormac has indicated he wants to rebuild the frontend in Python so that we can fix these errors ourselves.
5. This is fixed in the PR that rewrites the LCOE code.

With these new instructions, would you be able to retest?

Xiaofan1999:
Hi Femke,

Thanks for the updates, I retested the revised instructions.

1. The environment creation via `conda env create -f environment.yml` now works smoothly on my side.

2. After activating the environment, starting the backend from an activated Anaconda Prompt using `python Backend_FTT.py` works as expected. Adding Python to PATH did not change the behaviour of double-clicking the .cmd launcher on my setup, but launching from an activated conda environment is reliable.

3. FTT-H still fails (ValueError: Worksheet named 'HMAN' not found)

PR #291, opened 2025-12-23
Refactor levelised costs

Femkemilene:
This makes the levelised cost calculation about 5x faster, hopefully also clearer. It uses two types of levelised cost functions:

1. A simple one for when the purchase happens at t=0 (freight, heat, transport)
2. A more complicated one when there are multiple build years (power, steel)

Not implemented for freight yet, as I'm waiting for @AmAkther to do the PR for the major update.

Carlos: your tasks are:
1. To test if the code runs on your side.
2. To say if you understand the new logic, or whether it needs a better explanation
3. To test if the implementation is sufficiently similar to before. There are a few changes, as some of the accounting for standard deviations was wrong. Instead of adding all variance parts first, we did it in batches, overestimating uncertainty. I've corrected the original functions so the two functions correspond better. There is a Taylor approximation, which causes small deviations
4. To critique the implementation: is it sufficiently DRY (non-repetitive), are the docstrings clear, is the style okay.

### DIFF
--- a/SourceCode/ftt_core/ftt_get_levelised_costs.py
+++ b/SourceCode/ftt_core/ftt_get_levelised_costs.py
@@ -13,7 +13,7 @@ def get_levelised_costs(upfront, upfront_policies, upfront_sd,
                         lifetimes, r
                         ):
     """
-    Calculate levelised costs per unit with and without policy,
+    Calculate levelised costs per service delivered (e.g. Eur/kWh) with and without policy,
     and their standard deviation
     
     See https://www.wikiwand.com/en/articles/Annuity_(finance_theory)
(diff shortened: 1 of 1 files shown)


cadupsg (commented):
1. Yes, the code runs on my side.

2. Yes, I understood the new logic. It is well explained in the script. It requires the reader to have a previous understanding of LCOE and finance math, but the implementation itself is understandable by the logic and comments.

3. Yes, results from new implementation seem in line with the results of the old implementation, especially for mean LCOE. There are slightly larger variations for SD, which makes sense with the new calculation. Still, not a big difference in value.

4. The implementation looks DRY and readable, it is not hard to go along the code and understand each part, mainly with the separation between the base case and build-time case. Docstrings are very helpful, clear and don't disturb the flow of the code. It might be helpful to also add the units in the docstrings or write the formula with the units. I did not test with the right mid_year_correction (-0.5). If you are going to delete the old line when the code is ready (TODO), it is ok, but if you are keeping the "-1" option as a commented line, maybe it is worth adding a comment better explaining this part and why (-0.5) is being adopted.

PR #293, opened 2026-01-09
Merge cascading tipping points

cormacmlynch:
This pull request brings changes from the Cascading Tipping Points branch into main. 

 A local merge was done by @AmAkther. Some unneeded CSVs were deleted. 

All models (excl. industrial heat -- not tested) should be able to run independently and in combination with one another.


Femkemilene:
As this is code I wrote a lot of, I will take the responsibility of getting it over the line and responding to other people's comments. @cormacmlynch: can you be the official reviewer? If it doens't work in terms of timing, @AmAkther? There's another PR I want to do after this one (improving the levelised costs calculations), so I'd love to get this over the line fast to not hold the other PR up more than necessary.

 I've redone all the changes from the last half year that this PR seemed to revert. 

One question is how we want to implement the carbon tax. Currently, it's implemented separately for each sector. Is that the most logical implementation? Or do we want one or two carbon taxes and switches for which sectors fall under the carbon tax? The latter might be easier to implement. 

There's quite a bit of code here that I wrote in a 'quick and dirty' fashion for a single project, and I'm a bit worried about reducing overall code quality of the code. Keen to hear comments to improve code quality.

run_file.py
```diff
@@ -48,3 +50,9 @@
     # Fetch ModelRun attributes, for examination
     # Output of the model
     output_all = model.output
+
```
Femkemilene: Is this necessary? Rosie may have fixed this elsewhere

SourceCode/ftt_core/ftt_emissions_regulation.py
```diff
@@ -0,0 +1,504 @@
...
+        ice_indices.sort(key=lambda x: emissions_intensity[x], reverse=True)
+
+        current_emissions = get_fleet_emissions(sales, emissions_intensity)
+
+        # Iteratively redistribute until target met
+        while current_emissions > target_emissions and current_emissions > 0:
+            for ice_idx in ice_indices:
+                if sales[ice_idx] > 0:
+                    # Check for advanced counterpart
+                    adv_idx = find_advanced_counterpart(ice_idx, emissions_intensity)
```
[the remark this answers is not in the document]
ab-uoe: May work as intended in freight but may not work in transport

SourceCode/ftt_core/ftt_emissions_regulation.py
```diff
@@ -0,0 +1,504 @@
...
+                        sales[r, ice_idx, 0] -= transfer_amount
+                        sales[r, bev_index, 0] += transfer_amount
+
+                    # Recalculate fleet emissions
+                    current_emissions = get_fleet_emissions(sales[r, :, 0], em_int)
+
+                    if current_emissions <= target_emissions:
+                        break
+
+    return sales
```
ab-uoe: a suggestion would be to use an optimisation algorithm instead from a library since we are now not going to be constrained with uniformity with the fortran version. But maybe that could be a future development.

SourceCode/Freight/ftt_fr_lcof.py
```diff
@@ -3,189 +3,253 @@
...
+        
+        if not has_subsidies:
+            continue
+            
+        results[region] = {'subsidy_costs': [], 'tax_revenues': []}
+        print(f"\nChecking revenue neutrality for {region}")
+        
+        for veh_class in range(n_veh_classes):
+            bev_indices_class = [idx for idx in [31, 32, 33] if idx % n_veh_classes == veh_class]
+            ice_indices_class = [idx for idx in range(25) if idx % n_veh_classes == veh_class]
```
ab-uoe: if we can find a way to indentify indices more generally, rule based such as emissions==0, then this can be called/used for different FTTs

SourceCode/Freight/ftt_fr_lcof.py
```diff
@@ -3,189 +3,253 @@
...
+    """Calculate feebate rates for each vehicle class to maintain revenue neutrality"""
+    for r in range(len(titles['RTI'])):
+        for veh_class in range(n_veh_classes):
+            # Get indices for this vehicle class
+            bev_indices_class = [idx for idx in [31, 32, 33] if idx % n_veh_classes == veh_class]
+            ice_indices_class = [idx for idx in range(25) if idx % n_veh_classes == veh_class]
+            
+            # Calculate BEV subsidies for this class
+            bev_costs = data['BZTC'][r, bev_indices_class, c6ti['1 Purchase cost (USD/veh)']]
+            bev_numbers = data['ZEWI'][r, bev_indices_class, 0]
```
ab-uoe: consider using bev_units_sold instead

SourceCode/Freight/ftt_fr_main.py
```diff
@@ -77,273 +80,234 @@ def solve(data, time_lag, iter_lag, titles, histend, year, domain):
...
+    # Technology to fuel user conversion matrix
     zjet = data['ZJET'][0, :, :]
-    emis_corr = np.ones([len(titles['RTI']), len(titles['FTTI'])])
-
-    if year <= histend["RVKZ"]:
-
-        # U (ZEWG) is number of vehicles by technology
-        data['ZEWG'][:, :, 0] = data['ZEWS'][:, :, 0] * data['RFLZ'][:, np.newaxis, 0, 0]
+    # Initialise the emission correction factor
+    emis_corr = np.zeros([num_regions, num_techs])
```
ab-uoe: maybe emissions_correlatioon_factor

SourceCode/Freight/ftt_fr_main.py
```diff
@@ -77,273 +80,234 @@ def solve(data, time_lag, iter_lag, titles, histend, year, domain):
...
+            # Correction to market shares for each vehicle class
+            # Sometimes historical market shares do not add up to 1.0
+            for veh_class in range(n_veh_classes):
+                if (~np.isclose(summed_zews[r, veh_class, 0], 1.0, atol=1e-9)
+                    and summed_zews[r, veh_class, 0] > 0.0):
+                        data['ZEWS'][r, :, 0] = np.divide(data['ZEWS'][r, veh_class::n_veh_classes, 0],
+                                                    summed_zews[r, veh_class, 0])
+            
+        
+        # Calculate number of vehicles per technology. First reshape rflz into right format
```
ab-uoe: add maybe explanation saying that rflz is (regions, classes, 1) and we need (regions, techs, 1)

SourceCode/Freight/ftt_fr_regulatory_policies.py
```diff
@@ -0,0 +1,71 @@
...
+        reg_vs_exog = ((zwsa[r, :, 0] / no_it + endo_capacity[r]) 
+                      > zreg[r, :, 0]) & (zreg[r, :, 0] >= 0.0)
+     
+        # ZWSA is yearly capacity additions. We need to split it up based on the number of time steps, and also scale it if necessary.
+        dUkTK =  np.where(reg_vs_exog, 0.0, zwsa[r, :, 0] / no_it)
+    
+        # Correct for regulations due to the stretching effect. This is the difference in capacity due only to rflt increasing.
+        # This is the difference between capacity based on the endogenous capacity, and what the endogenous capacity would have been
+        # if rflz (i.e. total vehicles) had not grown.
+        dUkREG = -(endo_capacity[r] - endo_shares[r] * Utot[r]) \
```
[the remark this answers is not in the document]
Femkemilene: You're right. It should be Utot from previous timestep.

SourceCode/Freight/ftt_fr_main.py
```diff
@@ -353,50 +319,103 @@ def solve(data, time_lag, iter_lag, titles, histend, year, domain):
...
                     data_dt[var] = np.copy(data[var])
-
         
         # Calculate total investment by technology in terms of truck purchases
-        data['ZWIY'][:, :, 0] = data['ZEWY'][:, :, 0] * data['ZCET'][:, :, c6ti['1 Price of vehicles (USD/vehicle)']] * 1.263
-
+        data['ZWIY'] = data['ZEWI'] * data["BZTC"][:, :, c6ti['1 Purchase cost (USD/veh)'], None]
+        
+        if year == 2050:
+            print(f'Number of electric trucks globally: {np.sum(get_class(data['ZEWK'], 3)[:, 6])/1e6:.1f}M')
```
ab-uoe:
```suggestion
            print(f"Number of electric trucks globally: {np.sum(get_class(data['ZEWK'], 3)[:, 6])/1e6:.1f}M")
```

ab-uoe (changes requested):
Looking at US Nuclear generation 2022 -> 2023, the total generation is virtually unchanged but the sources is reshuffled. There could be a classification problem, specifically projections might be using a different classification than base year. The capacity factor for US nuclear stays locked at exactly 85.6% across the jump, confirming that generation and capacity were reset together. Global pumped hydro generation falls from 359,495 GWh in 2023 to 1,537 GWh in 2024 which is a 99.6% collapse; this might be part of the previous problem or the same problem in Freight below (bus and 3wheelers zeroing out). I think if I check other countries, the same structural break from historical to projection exists. 

Nuclear power also jumps 124% in a single year in UAE 2041 to 2042. 

There is also a steep decline in offshore wind in China, Germany, France and USA. Onshore to offshore ratio in 2050 is around 79:1. 

I remember that there was something about CCS technology that the model doesnt do well? since there is a near total absence of CCS generation through 2050. 

CCGT doesnt seem realistic in "58 NorthAfrica OPEC";  which continues to peak through 2050

FTT-Tr; is there no data for "rest of latin america"? zero ev adoption through 2050. 

FTT-Fr; Diesel HDT drops massively in China but seem to not move anywhere else which might not be intended as it assumes that no other country adopts or learns from Chinese market. Also was it intended that bus and three-wheeler shares zero out globally from 2023 to 2024?

Femkemilene:
Looking at US Nuclear generation 2022 -> 2023, the total generation is virtually unchanged but the sources is reshuffled. There could be a classification problem, specifically projections might be using a different classification than base year. The capacity factor for US nuclear stays locked at exactly 85.6% across the jump, confirming that generation and capacity were reset together. Global pumped hydro generation falls from 359,495 GWh in 2023 to 1,537 GWh in 2024 which is a 99.6% collapse; this might be part of the previous problem or the same problem in Freight below (bus and 3wheelers zeroing out). I think if I check other countries, the same structural break from historical to projection exists.

--> Nuclear mostly fixed. The remaining issue will be fixed with a PR that has been in draft mode since August waiting for this PR. For pumped hydro: I've turned it off completely now (it's a storage type, and we don't have a good way to implement this yet). 

Nuclear power also jumps 124% in a single year in UAE 2041 to 2042.
--> This is likely to be caused by solar being so dominent. I imagine some of our work on value factors will eventually solve this.

There is also a steep decline in offshore wind in China, Germany, France and USA. Onshore to offshore ratio in 2050 is around 79:1.
--> I've now included new gamma values, reducing the ratio to about 1:25. I had to tweak manually, which might point to a weakness in the gamma value automation, or just the exceptional year 2023 was for offshore (MWKA values are ignored in the gamma value automation). 

I remember that there was something about CCS technology that the model doesnt do well? since there is a near total absence of CCS generation through 2050.
--> You're absolutely right. Any technology that's not yet diffused struggles to diffuse in the model. In a future PR, we might change this behaviour to ensure that CCS can be implemented as a refurbishment. 

CCGT doesnt seem realistic in "58 NorthAfrica OPEC"; which continues to peak through 2050
--> I think this can be true: they have very cheap gas. 

FTT-Tr; is there no data for "rest of latin america"? zero ev adoption through 2050.
--> FTT:Tr only has 59 world regions. In future work linking FTT to MSET, we will have to make some changes here. That is: ensure the model can populate by proxy any number of world regions. For this to work, we need to separate true data from proxy data (all of the models have this mixed now). Future PR too.

FTT-Fr; Diesel HDT drops massively in China but seem to not move anywhere else which might not be intended as it assumes that no other country adopts or learns from Chinese market. Also was it intended that bus and three-wheeler shares zero out globally from 2023 to 2024?
--> There is work to enable cross-country diffusion of technologies. Again, requires good thinking so a separate PR and more work.

Femkemilene:
The code has become incredibly slow. I've tested whether this is due to a missing feature from main (locally merged main into the branch, will push soon). This is not the case, and there is simply some regression in performance of FTT:H. Instead of taking a couple of seconds, it now takes over 100s. Will investigate further. 

**Now fixed, at least mostly. The new currencyconversion is still really slow (>.5s for one call). But we're only calling it once now.**

SourceCode/Heat/ftt_h_main.py
```diff
@@ -49,7 +54,8 @@
 # -----------------------------------------------------------------------------
 # ----------------------------- Main ------------------------------------------
 # -----------------------------------------------------------------------------
-def solve(data, time_lag, iter_lag, titles, histend, year, domain):
+@profile
```
ab-uoe, line 57:
```suggestion
#@profile
```

ab-uoe (approved):
Comparing with previous results, the ratio between off and onshore wind is definitely better. However, US shows a decline in offshore wind 710 to 94 GWh. I'm not sure if this is intended. The UAE problem is actually already fixed under current values, might be good to check what fixed it. 

Just to clarify, have you seen the bus and three-wheel vehicle problem I mentioned previously? this I guess is another pull request. 

But overall I think the results are considerable better.

Femkemilene:
Many of the countries have no data for buses and threewheelers, so that's expected behaviour too. Thanks for the detailed review. I've put the gamma value negative for the US too.

PR #302, opened 2026-04-30
Set up 3 tests

Femkemilene:
To ensure that:
-> There's no nans in the output
-> There's no infs in the output
-> The runtime does not drop by a factor 2

If you want to test out the inf test, you can use the MWIY variable in ftt-power. That variable isn't impacting others, so you can do an isolated test without crashing the rest of the model.


Femkemilene:
@copilot: why did this fail?

bm638 (commented):
I ran the integration test on my own device and ran into some problems. This is the copilot summary of my findings. All three tests worked when I set included modules to "all". 

	Running the integration test with a reduced module set caused consistent NaNs originating from MSSE and MLSE being zero. These variables are required downstream in ftt_p_rldc and ftt_p_lcoe, but are not initialised unless all modules are enabled.
	This indicates that the current module system has implicit dependencies that are not enforced or documented. While enabling "ALL" modules resolves the issue, it means:
		○ The model is not robust to partial module execution
		○ The enable_modules configuration is misleading, as some modules are effectively mandatory
		○ Failures manifest as numerical instability (NaNs) rather than clear configuration errors
	For robustness, one of the following should be addressed:
		1. Explicitly define required module dependencies (e.g. power module depends on storage/RLDC)
		2. Add validation checks for critical variables (e.g. non-zero MSSE/MLSE)
		3. Fail early with informative errors when required state is missing
In the meantime, the integration test should use "ALL" modules to reflect the current operational assumptions of the model.

Femkemilene:
@bm638: in general for reviews (and bug reports), you should try to be as concrete as possible in making issues reproduceable. Which combination of modules leads to this behaviour? Which variables are NaNs? It's a bug if there are still dependencies between mouldes; this should have been fixed in a previous PR, but apparently not quite. In other words, Copilot is wrong in terms of solutions here and we humans still have a role to play :).

Femkemilene:
MSSE is the short-duration storage efficiency. As far as I'm aware, it's only used in the power sector. It should be 0 and not give any trouble in runs without the power sector, and should have a value if the power sector is turned on. I'll need more information to reproduce this.

bm638:
MSSE and MLSE are not being computed at all in the failing configuration.
	They are derived inside the RLDC/storage part of the power module (ftt_p_rldc.py), but that computation path is not being triggered when running:
	
	
	Python
	enable_modules = "FTT-Tr, FTT-P, FTT-H, FTT-Fr"
	
	In this configuration:
		• FTT-P (power) is enabled
		• but at least one upstream dependency required for storage/dispatch is not
	This leads to:
	MSSE = 0 everywhere
	MLSE = 0 everywhere
	
	which then causes divide-by-zero in ftt_p_rldc.py, e.g.:
	
	Python
	total_output / data['MSSE'][r, 0, 0]
	
	This produces NaNs that propagate into MEWL and then into LCOE (METC), where the test fails.
	
	Concretely:
		• First NaNs appear in MEWL
		• Then CF_mu becomes NaN
		• Then METC contains NaNs (triggering the exception)
		
	When switching to:

	Python
	enable_modules = "ALL"
	
	the issue disappears — MSSE and MLSE are populated with non-zero values and no NaNs occur.

Femkemilene:
enable_modules = "ALL" is not a valid input. Maybe the integration tests doesn't break as all output is likely 0 if you try this: no model runs. The standard set-up with all core models enabled is: enable_modules = "FTT-Tr, FTT-P, FTT-H, FTT-Fr", which works well on my side. The only thing I can think of is an issue with the environment setup being different between our two version? When copilot starts being wrong, it usually continues to be wrong, so best to stop using it.

bm638:
Oh that is embarrassing. I'm struggling at this stage of my learning to distinguish between what is helpful and what is wrong, especially as my knowledge of FTT code and variables is very shallow. Possibly an environments issue then? Some guidance in our catch up would be very helpful, thanks

Femkemilene:
The test should now fail if it's using 'ALL' or other invalid inputs for the enable_modules (or if we change how modules are read in). Apart from that, I'm happy with the tests now and will merge upon partial review

PR #307, opened 2026-05-14
New data loading

### LINKED ISSUE #6: Speed up import of input files
From profiling, it seems that 67% of run time is from importing input csv files. We should be able to speed this up with joblib, a library for parallellisation.

### LINKED ISSUE #320: Delete S3 scenario

cormacmlynch:
This pull request restructures how data is stored and read in to the model. The main changes are in input_functions.py.

Excel-based master files and per region per variable csv files have been replaced by a single csv file for each input variable. The benefits of this method are enhanced speed of data loading (no need for the xlsx to csv step and significantly fewer files to read) and better machine-friendly data format (for version control). The [Polars ](https://pola.rs/)library is also used in place of Pandas for [speed gains](https://towardsdatascience.com/polars-vs-pandas-an-independent-speed-comparison/).

The new data loading is only implemented for the power, heat, transport, and freight models. Data files for just S0 are included for now. Scenarios work as they did before -- a full set of variable csvs is not required for a scenario, only files different from S0 (e.g., MEWR in a scenario with regulation on coal).


Femkemilene:
@copilot: can you review all the .py files? No need to look at the file changes in the Inputs folder

Femkemilene:
@copilot: can you explain what you did here and why? (1) removed sys.path mutation and restored package import in SourceCode/support/convert_masterfiles_to_csv.py

Femkemilene:
@copilot: I've recently merged a PR into main. Can you merge main into this branch, so it's easier to assess that everything works together before I start reviewing the PR?

Femkemilene:
@cormacmlynch: In the previous version, we had a lot of error messages when things went wrong. Now, the data_loading was more finicky before, so maybe it's not as urgent, but I was wondering if it was a conscious choice to get rid of this functionality. Otherwise, it's lovely. I've tidied up slightly and hopefully increased the stability of the GUI (it gave a weird warning that Claude fixed). 

I think the type of data errors we need to check for:
1. Missing data (?) 
2. Incorrection dimensions (too many or too few rows and columns). 

I think that should be fairly clean to check for. My first attempt with Claude led to low-quality vibe code, so have not committed this. Would you have time to reintroduce these two checks?

Femkemilene:
The howto file on how to add new variables needs updating (docs / source / howto). Can the bullet point on the Excel file additions simply be removed or is there something new people need to take into account when adding new variables?

SourceCode/support/data_loading.py
```diff
@@ -0,0 +1,413 @@
...
+        Variable name (used to look up dims and forstart).
+    dims : dict
+        ``dims[var]`` → list of 4 dimension name strings.
+    titles : dict
+        Classification titles dict.
+    timeline : list of int
+        All simulation years.
+    tl_idx : dict
+        Pre-built ``{year: timeline_index}`` lookup.
+    forstart : dict
```
Femkemilene, line 246: This variable name (forstart or forecast start in VariableListing.csv) is something I've got confused at in my recent tipping_point branch changes. For variables like MEWG, TEWK, the year stated is 2010 and 2017, which is before the forecast start. Should this have a different name? Can't think of a good one, but if you set this value after the historical years, the historical values are being ignored.

SourceCode/support/data_loading.py
```diff
@@ -0,0 +1,444 @@
...
+
+    return data
+
+
+# ---------------------------------------------------------------------------
+# Internal helpers
+# ---------------------------------------------------------------------------
+
+def _wide_column_axis(variable_dims):
+    """Return the array axis (1, 2, or 3) used as the wide column axis.
```
Femkemilene, line 200: Could you explain what the wide column axis means in this function?

cormacmlynch: The function returns which of the dimensions is in the csv columns. Its usually time of course but can be cost matrix titles too and sometimes other titles.

Femkemilene:
@copilot resolve the merge conflicts in this pull request

Femkemilene:
@copilot : the gui seems to have broken when we went from SourceCode to ftt_source. Can you fix this?

Femkemilene (approved):
The new data loading is robust (fails when not in the right shape), the code is cleaner and the code is much faster. Thanks Cormac!

PR #317, opened 2026-06-05
Change ftt for package coupling w mset

ADanneaux:
Change FTT code to enable coupling with MSET using the FTT package.
This includes taking inputs from MSET as well as flexibility in reading setting files.
Core modifications include the need to make the typically 22 technology / 71 regions FTT more flexible.
MSET uses 12 technologies (open source OWID data) and a flexible number of regions.
The core issue is to remove hardcoding as much as possible and read from csv files for mapping. Default csv files are committed along with the code, but when coupling to another model, it is possible to define other input files.


Femkemilene:
@ADanneaux: did you test that the changes in code give the same output as before? And what tests did you do?

@copilot: could you rename the variable solar_t2ti to solar_tech_idx, solar_erti to solar_res_idx and oil_jti to oil_fuel_idx? Including for other technologies, resources and fuel types? That should make it more human readable.

Femkemilene:
@copilot resolve the merge conflicts in this pull request

Femkemilene:
Overall, this PR contains a lot of changes that are perfectly fine. Fewer elements are hard-coded. I struggle with a few:
1. The main power file now contains a large new function that is irrelevant to the run without MSET. This should be in a separate file
2. The settings.ini file now contains a lot of information. Is that the logical place to put this?
3. The increased emphasis on low-quality variable names such as t2ti. Let's see if I can reduce the complication here..

ftt_source/Power/ftt_p_integration_costs.py
```diff
@@ -75,40 +75,35 @@ def add_vre_integration_costs(data, titles):
...
         total_integration_cost = balancing + grid_costs
 
-        # Add integration costs to METC (generalized costs used in shares)
-        # This makes solar/wind more expensive as their share grows
-        # METC is in log space, so we add to the underlying cost
-        if solar_share > 0 and data['METC'][r, solar_idx, 0] != 0:
-            # Convert from log, add cost, convert back
-            current_cost = np.exp(data['METC'][r, solar_idx, 0])
-            data['METC'][r, solar_idx, 0] = np.log(current_cost + total_integration_cost)
+        # METC is in log space; convert out, add cost, convert back.
```
Femkemilene: METC is not in log space, that's something from the transport model only.

Femkemilene:
@ADanneaux: could you commit the LAT and RLDC files. That seems to be the thing that makes the check fail. You may have to adjust elements in the .gitignore file if these two files are not showing up to be committed.

ftt_source/Power/ftt_p_costc.py
```diff
@@ -21,8 +21,148 @@
...
+import pandas as pd
+
+# Local library imports
+from ftt_source.paths import get_utilities_path
+
+
+# %% tech-to-resource mapping and related lookups
+# -----------------------------------------------------------------------------
+
+def get_tech_to_resource(titles):
```
Femkemilene, line 35: All these calls are quite expensive. Reading in data takes ages. I don't know if this will fail the regression test as I cannot run the code without the missing LAT and RLDC file, but it's might be better to load this once.

ftt_source/Power/ftt_p_costc.py
```diff
@@ -21,8 +21,148 @@
...
+    df = pd.read_csv(csv_path, keep_default_na=False)
+
+    t2ti = list(titles['T2TI'])
+    erti = list(titles['ERTI'])
+    tech_resource_map = dict(zip(df['T2TI'], df['ERTI']))
+
+    return [erti.index(tech_resource_map[tech]) for tech in t2ti]
+
+
+def get_erti_jti_map(titles):
```
Femkemilene: Rename as "get_resource_to_fuel_map". At some point, we'll get rid of erti as a variable name. in the meantime, let's keep it simple for readers.

ftt_source/Power/ftt_p_costc.py
```diff
@@ -21,8 +21,148 @@
...
+    """
+    csv_path = get_utilities_path() / 'titles' / 'T2TI_CF_multiplier.csv'
+    df = pd.read_csv(csv_path, keep_default_na=False)
+
+    t2ti = list(titles['T2TI'])
+    return {t2ti.index(row['T2TI']): float(row['CF_multiplier'])
+            for _, row in df.iterrows()}
+
+
+def _detect_power_schema(t2ti):
```
Femkemilene, line 88: Does this mean we're hardcoding that there's only two options? I prefer our code to be flexible for new classifications too. For instance, Carlos is building a system to add rooftop solar.

ftt_source/Power/ftt_p_costc.py
```diff
@@ -284,55 +422,58 @@ def update_investment_cost(BCET, BCSC, CSC_Q, MEPD, MERC, tech_to_resource, inve
...
 
     # Uranium
-    MEPD[:, 0, 0] = MEWG[:, 0, 0] / BCET[:, 0, 13] * 3.6e-3
-    # Oil
-    MEPD[:, 1, 0] = MEWD[:, 2, 0] + MEWD[:, 3, 0] + MEWD[:, 4, 0]
-    # Coal
-    MEPD[:, 2, 0] = MEWD[:, 0, 0] + MEWD[:, 1, 0]
-    # Gas
-    MEPD[:, 3, 0] = MEWD[:, 6, 0]
+    n_t, n_e = gen_tech_indices['nuclear']
```
Femkemilene: n_t and n_e etc are cryptic variable names. Please spell out

ftt_source/Power/ftt_p_main.py
```diff
@@ -679,7 +739,8 @@ def solve(data, time_lag, titles, histend, year, domain):
             for var in vars_to_copy:
                 data_dt[var] = np.copy(data[var])
         
-        data = get_marginal_fuel_prices_mewp(data, titles, Svar)
+        if not mset_coupling:
```
Femkemilene, line 586: Interesting, so electricity costs are not coupled back to MSET?

Femkemilene:
@copilot : why is the output slightly different from main? This should just be a refactor when run without mset.

ftt_source/Power/ftt_p_costc.py
```diff
@@ -284,55 +422,58 @@ def update_investment_cost(BCET, BCSC, CSC_Q, MEPD, MERC, tech_to_resource, inve
...
-                    * 3.6e-3 # +  MEWD[11, :]   +   MEWD[10, :]
-    RERY[:, 4, 0] = MEPD[:, 4, 0]       # In PJ
+    b_ts, b_e = gen_tech_indices['biomass']
+    MEPD[:, b_e, 0] = sum(MEWG[:, t, 0] / BCET[:, t, 13] for t in b_ts) * 3.6e-3
+    RERY[:, b_e, 0] = MEPD[:, b_e, 0]
     # Biogas (which comes mostly from waste) (From here onwards cost curves are in TWh)
-    MEPD[:, 5, 0] = (MEWG[:, 4, 0] / BCET[:, 4, 13] 
-                     + MEWG[:, 5, 0] / BCET[:, 5, 13] ) * 3.6e-3
-    RERY[:, 5, 0] = MEPD[:, 5, 0]        # In PJ
+    bg_ts, bg_e = gen_tech_indices['biogas']
```
Femkemilene: Same: use longer variable names

ftt_source/Power/ftt_p_rldc.py
```diff
@@ -283,10 +275,12 @@ def rldc(data, MEWDt, time_lag, data_dt, year, t, titles, histend):
...
     # Bool to indicate which tech is VRE and which is not
     Svar = data['MWDD'][0, :, 5]
-    Snotvar = 1 - data['MWDD'][0, :, 5]
+    S_not_var = 1 - data['MWDD'][0, :, 5]
 
     # Sector coupling: update battery costs based on cumulative capacity
-    data = battery_costs(data, time_lag, year, t, titles, histend)
+    # Only when FTT:Transport is coupled (SCA dimension present AND sector_coupling enabled)
+    if 'SCA' in titles and sector_coupling:
+        data = battery_costs(data, time_lag, year, t, titles, histend)
```
Femkemilene, line 288: What happens to battery costs if either of these are not enabled? Can use a comment.

settings.ini
```diff
@@ -2,8 +2,18 @@
...
+storage_learning_base_year = 2022
+prsc_base_year = 2013
+ex_base_year = 2013
+rldc_start_year = 2013
+costc_skip_until = 2016
+usd_exchange_region = 34 USA (US)
+bcet_copy_range_end = 22
+carbon_price_conv_factor = 1.3281
+gamma_mode = multiplicative
+sector_coupling = True
```
Femkemilene, line 13: I wouldn't want all this information added to the settings file: I can imagine people may want to change the gamma mode in the period of phasing out the old additive ones, and sector coupling should also be a setting, but the details of how we initialise the base year of cost curves and residual load duration curves, that's a detail which does not matter much.

ftt_source/Power/ftt_p_main.py
```diff
@@ -129,29 +182,28 @@ def solve(data, time_lag, titles, histend, year, domain):
...
 
     # Initialisation
-    if year == 2013:
-        data['PRSC13'] = np.copy(data['PRSCX'])
-        data['EX13'] = np.copy(data['EXX'])
-        data['REX13'] = np.copy(data['REXX'])
+    if year == prsc_base_year:
+        data[prsc_var] = np.copy(data['PRSCX'])
+        data[ex_var] = np.copy(data['EXX'])
+        data[rex_var] = np.copy(data['REXX'])
```
Femkemilene: I'm not entirely sure what's happening here: are these variables called differently in MSET? If so, do they have a more human-readable name? This seems to be only changing the variable names on the left-hand side, even though they should be the same

ftt_source/Power/ftt_p_main.py
```diff
@@ -210,7 +265,7 @@ def solve(data, time_lag, titles, histend, year, domain):
             data['PRSC15'] = np.copy(data['PRSCX'])
 
 
-        # Set starting values for marginal costs of resources (MERC)
+        # Set starting values for marginal costs of resources (matches default FTT run)
```
Femkemilene: Ensure explanation of cryptic variable names are kept (or change variable names)

ftt_source/Power/ftt_p_main.py
```diff
@@ -219,7 +274,7 @@ def solve(data, time_lag, titles, histend, year, domain):
         data['MERC'][:, 7, 0] = 0.001
 
 
-        if year > 2013: 
+        if year > prsc_base_year:
```
Femkemilene: Why would this start with the PRSC baseyear? The 2013 is simply the initialisation start, so this might require a better variable name instead.

ftt_source/Power/ftt_p_main.py
```diff
@@ -236,25 +291,26 @@ def solve(data, time_lag, titles, histend, year, domain):
 
         # If first year, get initial MC, dMC for DSPCH ( TODO FORTRAN??)
         if not time_lag['MMCD'][:, :, 0].any():
-            time_lag = get_lcoe(data, titles)
+            time_lag = get_lcoe(data, titles, gamma_mode)
```
Femkemilene, line 185: Put in documentation that the old additive gammas will be removed at some point.

Femkemilene (commented):
Further comments

Femkemilene:
@copilot resolve the merge conflicts in this pull request

ftt_source/Power/ftt_p_main.py
```diff
@@ -58,34 +58,34 @@
...
+from ftt_source.Power.ftt_p_rldc import rldc
+from ftt_source.Power.ftt_p_early_scrapping_costs import early_scrapping_costs
+from ftt_source.Power.ftt_p_dspch import dspch, calculate_load_factors_from_dispatch
+from ftt_source.Power.ftt_p_lcoe import get_lcoe, set_carbon_tax
+from ftt_source.Power.ftt_p_fuel_price import get_marginal_fuel_prices_mewp
+#from ftt_source.Power.ftt_p_integration_costs import add_vre_integration_costs
+#from ftt_source.Power.ftt_p_surv import survival_function
+from ftt_source.Power.ftt_p_costc import cost_curves
+from ftt_source.Power.ftt_p_phase_out import set_linear_coal_phase_out
+from ftt_source.Power.ftt_p_initialisation import build_power_settings
```
Femkemilene, line 75: This is imported but never used. Does the import run the file? Or can this import be removed.

Femkemilene:
@ADanneaux: there are a few remaining comments. I've done the simple ones and ensured the front end works again, but would be good if the remaining are addressed. 

I wonder if the solution to avoid the superlong settings file is to strongly simplify the initialisation of the power sector model. I don't think we need that much historical detail. We might be able to split it into 3 time frames: first year (2010), remaining pre-MEWG years and true simulation. I've removed the erroneous currency conversions of the storage costs, which are in the same units as the rest of the model, and have corrected documentation about currency year (2023$ instead of 2013$). I don't know if those errors had an effect on coupling to MSET?

ftt_source/Power/ftt_p_main.py
Femkemilene: I've tried removing, but that does lead to an error, so keeping this for now. Maybe there are more elegant ways to do this coding-wise, but not important

Femkemilene:
I've strongly simplified the initialisation code. One of the reasons the settings file became so long was you / the MSET folks tried to move messy hard-coded numbers into the settings. Most of that old hard-coded mess was unnecessary or even wrong, so I've changed the initialisation logic. It does change some values, as the initialisation now works a bit differently, but not my much. 

@copilot: can you rereview the ftt_p_main.py changes?

Femkemilene (approved):

PR #323, opened 2026-06-18
Adjust the GUI results page preventing overflow

### LINKED ISSUE #314: Do not push down scenario overview for results when a large number of scenarios included
At the moment, the results tab in the frontend does not work as expected when you have many scenarios. It tries to display all, taking up so much horizontal space that the entire selection text is pushed down. That is: you cannot see any information anymore without scrolling down. This problem is evident when you have about 10-15 scenarios.

cormacmlynch:
Under model results loaded, only a max of 3 scenarios are shown for each model combination.

Also adjusted some of the properties of that container so that it should not 'jump' out of the tab.

Tested on both 24in monitor and 14in laptop.

### DIFF
--- a/GUI/page_results.py
+++ b/GUI/page_results.py
@@ -135,8 +135,11 @@ def render_results_page():
                                         for models_tuple, scen_list in sorted(
                                             models_by_scenario.items(), key=lambda x: (-len(x[0]), x[0])
                                         ):
+                                            # shorten long scenario lists for display if too many
+                                            if len(scen_list) > 3:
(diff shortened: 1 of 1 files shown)


Sanatparida01 (approved):
I have  tested this locally on a Mac with FTT-P, using 13 scenarios and an end year of 2050. The 'Model results loaded' summary correctly showed three scenario names followed by an ellipsis, while all 13 scenarios remained selectable. The layout stayed within the Load Files tab, and I successfully plotted the scenarios in Analysis. Everything worked as expected in my testing.

PR #326, opened 2026-06-19
Fix marginal costs

Femkemilene:
The calculations were indicating that storage levelised costs (including investment costs) had to be added to variable renewable costs, which gave higher marginal costs for renewables than for many fossil fuel costs, which is unrealistic. The marginal cost of storage is much lower. If / when we do storage as a separate tech, we can include this.


ldxib2 (approved):
Looks good, storage costs switch removed and var listing updated

PR #327, opened 2026-06-22
Fix double counting in marginal pricing

Femkemilene:
The old marginal pricing included a switch to VRE prices as baseload disappeared, but also an additional VRE weight as VRE shares increased. This was double counting. Instead, I've reverted back the the RLDC logic, but introduced a smoothing so that we don't have an abrupt regime shift when VRE takes over from baseload.

I've further increased the costs of the reserves and peak prices, as in practice these prices have been going up incredibly rapidly. This might still be too little.


pv-camecon (approved):
The code changes fix the double counting.

PR #329, opened 2026-06-24
Remember result dimension selections between variables

### LINKED ISSUE #324: Remember selection of region/technologies when switching variable
When you switch from one variable to another (MEWG, generation to MEWS, capacity shares) in the Results tab, you have to reselect the right countries and technologies. Instead:

- When variables have an overlap in dimensions (such as both having a country element), or both using the T2TI technology classification, this choice should be remembered instead. 
- The settings such as 'All' and 'Sum' should also be remembered when switching between variables with identical classification. 

I believe this an another good first issue, Sanat, and I imagine that you only need a few lines of code to change this behaviour. Pinging Cormac here, as he developed the new frontend, and may be able to review a PR to fix this behaviour.

Femke

Sanatparida01:
Fixes #324.

- This PR updates the Results page so that dimension selections are remembered by classification when switching between variables. For example, if two variables both use Region or the same FTT:P technology classification, the selected values are kept instead of resetting to the first available option.

- I tested this locally with FTT-P results. Switching between MEWG and MEWS preserved the selected regions and FTT:P technologies. I also tested the All/Sum options for matching classifications, and these were preserved when switching between variables with the same dimension setup.

- I additionally checked a variable with a different dimension setup, PRSCX. In that case, the selected regions were preserved, but the FTT:P technology selector was not shown, which is expected because PRSCX does not use that technology dimension.

- I also tried a few other variable switches to check that selections are only carried over when the underlying classifications match.

### DIFF
--- a/GUI/page_results.py
+++ b/GUI/page_results.py
@@ -316,8 +316,14 @@ def render_results_page():
         nonlocal result_type_selector, result_type_hint
         _baseline_hint = 'Select a baseline and ≥1 other scenario to enable comparison.'
         dimension_container.clear()
-        # Reset dimension selections when the variable changes to avoid stale
-        # position-based selections from a previous variable bleeding through.
+        # Store selections by classification so shared dimensions are remembered
+        # when switching variables, e.g. RTI region or T2TI technology.
+        if not hasattr(state, 'dim_selection_cache'):
+            state.dim_selection_cache = {}
+        if not hasattr(state, 'dim_aggregate_cache'):
+            state.dim_aggregate_cache = {}
+
+        # Reset position-based selections before rebuilding them for this variable.
         state.dim_selections = [{} for _ in range(4)]
         state.dim_aggregate = [False, False, False, False]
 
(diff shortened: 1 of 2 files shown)


Femkemilene:
@copilot: question only, do not make code changes. When I go to settings and come back to the graph, the line disappears. I only have to reselect one dimension for it to reappear, but what can I do to make that more sticky?

@Sanatparida01: I've tested it, and it's a great improvement. This is going to save users a lot of headaches. Thanks so much! Of the copilot suggestions, the first, third and last seem valid. Could you implement those (or ask copilot to implement if you're able to click on those too).

Femkemilene (changes requested):
See above

Sanatparida01:
@Femkemilene  Thanks for the review. I’ve implemented the first, third, and last Copilot suggestions:
---moved the dimension selection caches into GUIState,
---updated the All checkbox so it stays in sync with the selected values,
---simplified the repeated assignment in the All handler.

Femkemilene (approved):
All changes done

PR #336, opened 2026-07-08
Automatically select latest Results file

### LINKED ISSUE #330: Automatically select latest Results file in frontend
When you run a new (set of) scenarios in the frontend, you have to manually select that pickle file. It would be good if instead the latest Results file is selected automatically. That way, users can simply add other results files where desired.

Sanat, let me know if this is something you're keen on working. If you want something more challenging, let me know.

Sanatparida01:
Fixes #330.

This PR updates the Results page so the most recently modified pickle file in the Output folder is selected automatically when the Results page opens. This means that after running a new set of scenarios, the latest output file is loaded without requiring the user to manually select it first.

Users can still add other results files manually if they want to compare multiple outputs.

Tested locally by running FTT-P / S0 with new output names `sanatparida_test03` and `sanatparida_test04`. After each run, the latest output file was selected automatically in Results → Load Files, and I could still manually add another result file for comparison.

### DIFF
--- a/GUI/page_results.py
+++ b/GUI/page_results.py
@@ -55,7 +55,7 @@ def render_results_page():
                                 pickle_display_names = list(pickle_files_map.keys())
                                 latest_pickle_file = engine.get_latest_pickle_file()
                                 initial_result_files = state.selected_result_files
-                                if latest_pickle_file and latest_pickle_file not in initial_result_files:
+                                if latest_pickle_file and not initial_result_files:
                                     initial_result_files = [latest_pickle_file]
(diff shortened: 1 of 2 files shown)


Sanatparida01:
I’ve addressed the Copilot review comments:
- updated `get_latest_pickle_file()` to use `Path.stem` and skip files that cannot be stat’ed,
- changed the Results page auto-selection so the latest file is only selected when there is no existing selection.

Femkemilene (approved):
This is great. I've tested it all out, and it works as expected.

PR #337, opened 2026-07-20
Group multidimensional result lines by style.  

### LINKED ISSUE #334: Use different line styles for multiple comparisons
Sometimes you select multiple elements in multiple dimensions. For instance, Belgium and Denmark, and onshore wind and solar PV. This results in four lines. It would be neat if those lines are grouped:

1. Each country gets a separate colour
2. Each technology (or whatever the second dimension is) gets its own line style. Normal, dashed, dotted etc. 

When only one country is selected, and multiple technologies, I imagine the best route is to simply use multiple colours for the technologies. This should only trigger when there is a multidimensional selection.

Sanatparida01:
Fixes #334.

This PR updates the Results chart styling when multiple dimensions are selected.

When two or more dimensions have multiple selected values, the first selected dimension controls line colour and the second controls line style. For example, when selecting multiple countries and multiple technologies/resources, each country keeps a consistent colour while each technology/resource is shown with a different line style.

This makes multidimensional plots easier to read without changing the underlying data or calculations.

Tested locally:
- One country with multiple technologies/resources still uses normal separate colours.
- Two countries with two technologies/resources now groups by country colour and technology/resource line style.
- When one dimension is summed, the grouped styling switches off as expected.

### DIFF
--- a/GUI/results_engine.py
+++ b/GUI/results_engine.py
@@ -406,6 +406,13 @@ class ResultsEngine:
         ]
         return multi_dims[:2] if len(multi_dims) >= 2 else []
 
+    def _style_position(self, selected_indices, dim_idx, value_idx):
+        """Map a dimension value index to its position in current selection order."""
+        try:
+            return selected_indices[dim_idx].index(value_idx)
+        except ValueError:
+            return value_idx
+
     def _get_trace_line_style(self, combo, style_dims, selected_indices):
         """Return Plotly line styling for grouped multidimensional selections."""
         if len(style_dims) < 2:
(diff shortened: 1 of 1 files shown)


Femkemilene:
@copilot: can you explain why I'm not seeing any lines now in the frontend?

Femkemilene:
@copilot: the line style behaviour is odd. When you add a second technology (dim2), it starts with dashdot instead of dash, which doesn't look as calm as the simple dash. How come and can you fix?

Femkemilene:
@copilot: I think the same issue is present with the countries. If I add country 1, 11 and 21, it's all the same colour rather than the first 3 colours. Also, was it necessary to be that wordy in the previous commit? The code is now difficult to read.

Femkemilene (approved):

Sanatparida01:
Thanks Femke. I understand the issue now , the initial implementation worked for the basic case I tested, but did not handle sparse selected indices well. The final selected order mapping is clearer and gives the expected colour/dash sequence. I will keep this edge case in mind for future dashboard testing.

PR #345, opened 2026-08-19
Add stop button for running simulations

### LINKED ISSUE #341: Include option to stop run from frontend
Sometimes you realise that you've started a simulation, but you've not included the right settings. E.g. you may have forgotten to add a model, or not selected the right scenarios. 

There should be a stop button next to the Run button, so that you can reselect scenarios without having to wait for all of the scenarios to be run. 

Relatedly, at the moment you can change the selection while a run is ongoing, but it does not appear to do anything.

Sanatparida01:
Fixes #341.

This PR adds a Stop button next to the Run button so users can cancel a simulation after it has started.

Changes:
- Adds a Stop button while a model run is active.
- Disables the model, scenario, end-year, and output-name controls during a run so users do not change settings that cannot affect the current run.
- Sends a stop request to the running model and cancels at a safe checkpoint.
- Re-enables the controls after the run is cancelled or completed.

Tested locally:
- The Stop button appears while a run is active.
- Clicking Stop logs “Stop requested” and “Run cancelled by user”.
- The controls become usable again after cancellation.

### DIFF
--- a/GUI/page_run.py
+++ b/GUI/page_run.py
@@ -90,12 +90,12 @@ def render_run_page():
     async def start_run():
         nonlocal update_timer, stop_event
         
-        # Disable run button and set running state (disables navigation buttons)
+        # Prepare cancellation state before making the Stop button clickable
         run_btn.disable()
-        stop_btn.enable()
-        set_inputs_enabled(False)
         state.is_running = True
         stop_event = Event()
+        set_inputs_enabled(False)
+        stop_btn.enable()
         progress_bar.set_value(0)
         log_console.clear()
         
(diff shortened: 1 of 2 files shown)


Sanatparida01:
Thanks for the review. I’ve addressed the Stop button ordering issue and replaced the cancellation magic string with a dedicated RunCancelledError exception.

I tested this locally . The Stop button appears while a run is active; clicking it logs “Stop requested” and “Run cancelled by user”, and the run is cancelled without crashing the frontend.

Femkemilene:
@copilot: to ensure that the order of comments is correct (i.e. 'Running model.. doesn't come after the cancellation message), an agent reintroduced an Runtimeerror again. Does this lead to a reintroduced fragility? Or are the new changes robust?

Femkemilene (approved):

PR #347, opened 2026-09-20
Issue run progress updates

### LINKED ISSUE #342: Better progress tracking
When you run a series of scenarios, or any type of run, the progress tracker doesn't really work that well. It indicates for the most time that there is no run progress and then suddenly it's fully done. 

Two options:
1. Improve the progress bar, using data about the year of the run
2. Improve printing, giving a print statement for each scenario that has been run

Sanatparida01:
## What changed

The progress bar now updates as each simulation year is completed. When multiple scenarios are selected, their progress is combined so the bar moves steadily from 0% to 100% across the whole run.

The run log also shows when each scenario starts and finishes, including the time taken for that scenario. This makes it easier to see what the model is currently doing.

The existing Stop button continues to work during the run.

## Testing

I tested the dashboard with scenarios S0, S2, and S3 through 2050. The progress bar moved after each year, reached 100% when all three scenarios finished, and the results were saved successfully.

Four automated tests also pass, covering progress updates, scenario messages, failures, and cancellation.

Closes #342


Femkemilene:
Thanks so much! 

The tests now lack a docstring. I'm not sure if we need that many tests and what they are for. The name of the file is not that informative (should say tests_progress_bar likely)

Femkemilene (changes requested):
I've tested this out, and the progress bar itself works really well. In terms of the other behaviour:
1. It would be good to only have one line per scenario, so that the comments in the dialogue don't become overwhelming. Running scenario X, instead of "start scenario X"
2. To me, total time is more useful than per-scenario timing.

